PR #1, opened 2020-01-16
Header Table

fosskers:
This PR uses the `BlockT` table type to ingest the Header Stream into a prototype DB.


lib/ChainwebDb/Types/Block.hs
```diff
@@ -1,59 +1,50 @@
...
-  , _block_hash :: C f DbHash
-  , _block_powHash :: C f DbHash
-  , _block_target :: C f DbHash
-  , _block_weight :: C f DbHash
-  , _block_epochStart :: C f UTCTime
-  , _block_nonce :: C f Word64
-  , _block_miner :: PrimaryKey MinerT f
-  } deriving Generic
+  { _block_id           :: C f Int
+  , _block_creationTime :: C f Int
```
mightybyte: What was the reason for switching this to `Int`?  I'd like to keep it as UTCTime if possible.

lib/ChainwebDb/Types/Block.hs
```diff
@@ -1,59 +1,50 @@
...
+  , _block_chainId      :: C f Int
+  , _block_height       :: C f Int
+  , _block_hash         :: C f DbHash
+  -- , _block_powHash      :: C f DbHash
```
mightybyte: I originally put this field in because this DB schema is supposed to be optimized for querying/analytics and I think the powHash is definitely something that people will want to do analysis on.  Is there any reason we can't put this in?

lib/ChainwebDb/Types/Block.hs
```diff
@@ -1,59 +1,50 @@
...
+  deriving anyclass (Beamable)
 
 Block
   (LensFor block_id)
   (LensFor block_creationTime)
   (LensFor block_chainId)
   (LensFor block_height)
   (LensFor block_hash)
-  (LensFor block_powHash)
+  -- (LensFor block_powHash)
```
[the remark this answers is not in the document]
mightybyte: We can do it without depending on `chainweb` the way the block explorer does it, which is to request both the JSON and binary representation and then calculate the powHash from the binary representation.

https://github.com/kadena-io/block-explorer/blob/master/frontend/src/Frontend/ChainwebApi.hs#L140

We should definitely do that.

lib/ChainwebDb/Types/Block.hs
```diff
@@ -1,59 +1,50 @@
...
   (LensFor block_target)
   (LensFor block_weight)
   (LensFor block_epochStart)
   (LensFor block_nonce)
-  (MinerId (LensFor block_miner))
+  -- (MinerId (LensFor block_miner))
```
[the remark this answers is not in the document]
mightybyte: We definitely also want to pull down the payloads and include this field.  That could come in a follow-on PR if it makes sense.  (Although that would also mean we starting getting into migrations sooner.)

lib/ChainwebDb/Types/Transaction.hs
```diff
@@ -1,44 +1,41 @@
...
+import ChainwebDb.Types.Block
 ------------------------------------------------------------------------------
 
 
 ------------------------------------------------------------------------------
 data TransactionT f = Transaction
-  { _transaction_id :: C f Int
-  , _transaction_chainId :: C f Int
-  , _transaction_block :: PrimaryKey BlockT f
+  { _transaction_id           :: C f Int
```
mightybyte: Let's avoid this vertical alignment for fields in a data type.  It makes diffs larger and increases the number of merge conflicts.

lib/ChainwebDb/Types/Transaction.hs
```diff
@@ -1,44 +1,41 @@
...
-{-# LANGUAGE GADTs #-}
-{-# LANGUAGE ImpredicativeTypes #-}
-{-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE TypeApplications #-}
-{-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE DeriveAnyClass       #-}
```
mightybyte: Same comment about alignment here.  The readability benefit for the language extension list just isn't worth the cost.

lib/ChainwebDb/Types/Block.hs
fosskers: The migration code really didn't like it. I can try to find a way back around it in future.

PR #14, opened 2020-02-18
Fix Backfill

fosskers:
This PR fixes `backfill` which was broken due to Postgres insert transactions interacting poorly with Haskell concurrency.

It also makes sure that the chains are backfilled (relatively) evenly. Previously the recursive nature of the code would fly down one chain and never look back.

### DIFF
--- a/exec/Chainweb/Backfill.hs
+++ b/exec/Chainweb/Backfill.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE LambdaCase #-}
 
 module Chainweb.Backfill ( backfill ) where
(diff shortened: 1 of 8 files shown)


exec/Chainweb/Backfill.hs
```diff
@@ -37,29 +40,54 @@ newtype Parent = Parent DbHash
...
+  b@(T2 _ (Parent p)) <- atomically $ readTBQueue toCheck
+  m <- runBeamPostgres c $ do
+    blk <- runSelectReturningOne $ lookup_ (blocks database) (BlockId p)
+    hum <- runSelectReturningOne $ lookup_ (headers database) (HeaderId p)
+    pure (void blk <|> void hum)
+  maybe (atomically $ writeTBQueue toLook b) mempty m
+
+-- | Lookup a parent, write it to the Work Queue, and ready /its/ parent for analysis.
+--
+-- This can be ran over multiple threads.
```
fosskers: Not actually true - the transactions still mix together somehow at a low level.

PR #19, opened 2020-02-25
Use `UTCTime`

fosskers:


lib/ChainwebDb/Types/Transaction.hs
```diff
@@ -78,3 +83,10 @@ instance Table TransactionT where
     deriving stock (Generic)
     deriving anyclass (Beamable)
   primaryKey = TransactionId . _tx_requestKey
+
+--------------------------------------------------------------------------------
+-- Orphan
+
+-- Until this is merged: https://github.com/tathougies/beam/pull/422
+instance HasDefaultSqlDataType Postgres UTCTime where
+  defaultSqlDataType _ _ _ = timestampType Nothing True
```
fosskers: This will eventually become a conflict.

PR #23, opened 2020-02-29
Store target and weight as numbers

mightybyte:


mightybyte:
The `target` and `weight` fields in the block table now get created with the right postgres types.  The listen command works, but backfill is failing with BeamRowReadError ColumnTypeMismatch for some reason that I haven't figured out yet.

PR #29, opened 2020-03-23
More careful Backfill

fosskers:
This PR prevents `backfill` from producing uneven results across chains.

### DIFF
--- a/exec/Chainweb/Backfill.hs
+++ b/exec/Chainweb/Backfill.hs
@@ -24,7 +24,7 @@ import           Database.Beam.Postgres (Connection, runBeamPostgres)
 ---
 
 backfill :: Env -> IO ()
-backfill e@(Env _ c _ v) = withPool c $ \pool -> do
+backfill e@(Env _ c _ v _) = withPool c $ \pool -> do
   putStrLn "Backfilling..."
   cont <- newIORef 0
   mins <- minHeights v pool
--- a/exec/Chainweb/Backfill.hs
+++ b/exec/Chainweb/Backfill.hs
@@ -24,13 +24,14 @@ import           Database.Beam.Postgres (Connection, runBeamPostgres)
 ---
 
 backfill :: Env -> IO ()
-backfill e@(Env _ c _ _) = withPool c $ \pool -> do
+backfill e@(Env _ c _ v) = withPool c $ \pool -> do
   putStrLn "Backfilling..."
   cont <- newIORef 0
-  mins <- minHeights pool
+  mins <- minHeights v pool
   let !count = M.size mins
   printf "Valid Chains: %d\n" count
-  when (count == 10) $ traverseConcurrently_ Par' (f pool cont) $ lookupPlan mins
+  when (count == length (chainsByVersion v))
+    $ traverseConcurrently_ Par' (f pool cont) $ lookupPlan mins
   where
     f :: P.Pool Connection -> IORef Int -> (ChainId, Low, High) -> IO ()
     f pool count range = headersBetween e range >>= \case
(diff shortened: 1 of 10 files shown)


exec/Chainweb/Backfill.hs
```diff
@@ -27,7 +28,9 @@ backfill e@(Env _ c _ _) = withPool c $ \pool -> do
   putStrLn "Backfilling..."
   cont <- newIORef 0
   mins <- minHeights pool
-  traverseConcurrently_ Par' (f pool cont) $ lookupPlan mins
+  let !count = M.size mins
+  printf "Valid Chains: %d\n" count
+  when (count == 10) $ traverseConcurrently_ Par' (f pool cont) $ lookupPlan mins
```
fosskers: Magic number.

mightybyte: We probably shouldn't have the constant `10` here.  It should get it from the node.

exec/Chainweb/Env.hs
```diff
@@ -54,6 +55,9 @@ newtype Url = Url String
 newtype ChainwebVersion = ChainwebVersion Text
   deriving newtype (IsString)
 
+chainsByVersion :: ChainwebVersion -> [ChainId]
```
mightybyte: Let's get this from the node's `/info` endpoint instead of hard coding it.

exec/Chainweb/Backfill.hs
```diff
@@ -23,11 +25,14 @@ import           Database.Beam.Postgres (Connection, runBeamPostgres)
...
 backfill :: Env -> IO ()
-backfill e@(Env _ c _ _) = withPool c $ \pool -> do
+backfill e@(Env _ c _ _ cids) = withPool c $ \pool -> do
   putStrLn "Backfilling..."
   cont <- newIORef 0
-  mins <- minHeights pool
-  traverseConcurrently_ Par' (f pool cont) $ lookupPlan mins
+  mins <- minHeights cids pool
+  let !count = M.size mins
+  printf "Valid Chains: %d\n" count
```
mightybyte: Let's change this to display a more informative error message to the user.  Or, better yet...instead of requiring that the user wait until all chains are present, can we use the chain list that we are now retrieving from the node?

exec/Main.hs
```diff
@@ -23,12 +25,15 @@ main = do
...
     putStrLn "DB Tables Initialized"
     m <- newManager tlsManagerSettings
-    let !env = Env m pgc u v
-    case c of
-      Listen -> listen env
-      Backfill -> backfill env
-      Gaps -> gaps env
-      Single cid h -> single env cid h
+    allChains m u >>= \case
+      Nothing -> printf "[FAIL] Unable to contact %s for Chain counts\n" u >> exitFailure
```
mightybyte: While we're changing the  above, let's make this a little more accurate.  Something like `Unable to connect to node`.

exec/Chainweb/Backfill.hs
fosskers: (1) Yes, let's change the message.
(2) Those are slightly orthogonal things. `mins` is the list of minimum heights as determined by looking at the DB. We need some definite starting point for each chain, and it might be dangerous to guess

exec/Main.hs
fosskers: The `u` is the address of the Node, or is that still not clear enough?

mightybyte: Eh...I think the "for chain counts" is a little too specific.

exec/Chainweb/Backfill.hs
mightybyte: Ahh yes, I had a vague realization there might be something else needed.  A change to the message would be fine then.

PR #31, opened 2020-03-25
Print block rate, tweak names

mightybyte:
Showing the block fill rate gives us more insight into backfill performance and behavior.

### DIFF
--- a/exec/Chainweb/Backfill.hs
+++ b/exec/Chainweb/Backfill.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NumericUnderscores #-}
@@ -33,7 +32,7 @@ backfill :: Env -> IO ()
 backfill e@(Env _ c _ _ cids) = withPool c $ \pool -> do
   cont <- newIORef 0
   mins <- minHeights cids pool
-  let !count = M.size mins
+  let count = M.size mins
   if count /= length cids
     then do
       printf "[FAIL] %d chains have block data, but we expected %d.\n" count (length cids)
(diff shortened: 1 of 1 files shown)


exec/Chainweb/Backfill.hs
```diff
@@ -33,7 +32,7 @@ backfill :: Env -> IO ()
 backfill e@(Env _ c _ _ cids) = withPool c $ \pool -> do
   cont <- newIORef 0
   mins <- minHeights cids pool
-  let !count = M.size mins
+  let count = M.size mins
```
fosskers: Banging nearly everything in a `let` is a practice I adopted from Niklas Hambuchen to avoid inlining GOTCHAs. It's probably not the end of the world here, though.

exec/Chainweb/Backfill.hs
```diff
@@ -55,19 +54,20 @@ progress count mins = do
...
-    printf "[INFO] Progress: %d/%d blocks (%.2f%%). ~%d %s remaining.\n"
-      total remaining perc tilDone period
+    let perc = (100 * fromIntegral completed / fromIntegral total) :: Double
+        elapsedMinutes = (now - start) / 60
+        blocksPerMinute = (fromIntegral completed / realToFrac elapsedMinutes) :: Double
+        estMinutesLeft = floor (fromIntegral (total - completed) / blocksPerMinute) :: Int
+        (timeUnits, timeLeft) | estMinutesLeft < 60 = ("minutes" :: String, estMinutesLeft)
+                              | otherwise = ("hours", estMinutesLeft `div` 60)
+    printf "[INFO] Progress: %d/%d blocks (%.2f%%), ~%d %s remaining at %.0f blocks per minute.\n"
+      completed total perc timeLeft timeUnits blocksPerMinute
```
fosskers: Could `floor` to `blocksPerMinute` here and then use `%d` in the pattern.

exec/Chainweb/Backfill.hs
mightybyte: In this case it's forced on the next line by the conditional.  I tend to think that sprinkling bangs everywhere unnecessarily obscures the times when they're actually important.

exec/Chainweb/Backfill.hs
mightybyte: I tried that.  It required an extra type annotation, so I decided this was cleaner.

PR #43, opened 2021-03-31
Add events

sirlensalot:
Adds `events` table and provides the following new queries/endpoints:
  - `tx` which is a "tx detail" search by request key returning a single result, with full events
  - `events` which is an event search supporting the following params (along with limit and offset):
    - `param` for param text wildcard search
    - `requestkey` for exact request key search
    - `name` for event name wildcard search (e.g. "TRANSFER")
    - `idx` for exact event idx search (meaningless without `requestKey` but oh well)


mightybyte (commented):
Looks good!  I'll see if I can give it a try.  The only question mark I can think of right now is how it will behave on older blocks before events were introduced.  But that seems like it's probably an issue handled by chainweb-node.

mightybyte:
Do you want to expose any kind of REST API for querying this information?  As it stands, this PR will only populate the database.  If we want the chainweb-data server to expose any kind of querying functionality for this new table, that will have to be added here:

https://github.com/kadena-io/chainweb-data/blob/master/exec/Chainweb/Server.hs#L127

stack.yaml
```diff
@@ -15,7 +15,7 @@ extra-deps:
   - which-0.1.0.0
   - witherable-class-0
   - github: kadena-io/chainweb-api
-    commit: 5d345f46a9cb08e55a4a9f0be113756333d9ba7d
+    commit: cf8676802c3685b0500cedc2f250a65c7f2112e4
```
mightybyte, line 18: I don't see this commit anywhere in the chainweb-api repo.

lib/ChainwebDb/Types/Event.hs
```diff
@@ -0,0 +1,60 @@
...
+import Database.Beam
+import Database.Beam.Postgres (PgJSONB)
+------------------------------------------------------------------------------
+import ChainwebDb.Types.Transaction
+------------------------------------------------------------------------------
+
+
+------------------------------------------------------------------------------
+data EventT f = Event
+  { _ev_requestKey :: PrimaryKey TransactionT f
```
mightybyte, line 30: @sirlensalot I think this has problems for events that aren't attached to a transaction.  (The main event in question here is coinbase events.)  It seems like we need to modify this event table schema to allow this.  Any thoughts about a good way to do this?  One idea that comes to mind is adding a `blockHash` field and making the `requestKey` field a `Maybe`.

Another factor along these lines to consider is a mechanism to inherently track the progress of event backfill.  The other PR introduces a `numEvents` field into the transactions table to facilitate this, but we also need to consider the case where a block has no transactions (and thus no `numEvents` field) but we still need to track progress backfilling the coinbase events.  Any thoughts on this?

emmanueldenloye: I don't think the request key should be a maybe type. Coinbase events still have request keys within them. Unless there is something that I'm missing and there are other types of events without request keys?

exec/Chainweb/Backfill.hs
```diff
@@ -72,6 +87,57 @@ backfill e delay = do
...
+backfillEvents e args = do
+  cutBS <- queryCut e
+  let curHeight = fromIntegral $ cutMaxHeight cutBS
+      cids = atBlockHeight curHeight allCids
+  counter <- newIORef 0
+
+  missingTxs <- countTxMissingEvents pool
+  -- assume a default value suitable for mainnet... the value 11138000 is the
+  -- height at which events were activated in chainweb-node
+  missingCoinbase <- countCoinbaseTxMissingEvents pool (fromMaybe 1138000 $ _backfillArgs_coinBaseMinHeight args)
```
mightybyte, line 100: We should probably allow this to be passed in by the user.

mightybyte: Oh, duh...it is

exec/Chainweb/Backfill.hs
```diff
@@ -72,6 +87,74 @@ backfill e delay = do
...
+  where
+    delay = _backfillArgs_delayMicros args
+    coinbaseDelay = _backfillArgs_coinbaseDelayMicros args
+    pool = _env_dbConnPool e
+    allCids = _env_chainsAtHeight e
+    delayFunc =
+      case delay of
+        Nothing -> pure ()
+        Just d -> threadDelay d
+    eventsActivationHeight = fromMaybe 1138000 $ _backfillArgs_eventsActivationHeight args
```
mightybyte, line 127: If we're going to put a hard-coded default in here, I think we should case on the chainwebVersion and use the correct value depending on whether it's mainnet01 or testnet04.

exec/Chainweb/Backfill.hs
```diff
@@ -72,6 +87,74 @@ backfill e delay = do
...
+    f :: Map ChainId Int64 -> IORef Int -> ChainId -> IO ()
+    f missingCoinbase count chain = do
+      blocks <- getTxMissingEvents chain pool (fromMaybe 100 $ _backfillArgs_eventChunkSize args)
+      forM_ blocks $ \(h,(current_hash,ph)) -> do
+        payloadWithOutputs e (T2 chain ph) >>= \case
+          Nothing -> printf "[FAIL] No payload for chain %d, height %d, ph %s\n"
+                            (unChainId chain) h (unDbHash ph)
+          Just bpwo -> do
+            P.withResource pool $ \c -> runBeamPostgresDebug putStrLn c $
```
mightybyte, line 136: Let's get rid of these debugs.  They slow down the backfill significantly.

exec/Chainweb/Backfill.hs
```diff
@@ -88,3 +171,67 @@ minHeights cids pool = M.fromList <$> wither selectMinHeight cids
       $ orderBy_ (asc_ . _block_height)
       $ filter_ (\b -> _block_chainId b ==. val_ (fromIntegral cid))
       $ all_ (_cddb_blocks database)
+
+-- | For all blocks written to the DB, find the shortest (in terms of block
+-- height) for each chain.
+countTxMissingEvents :: P.Pool Connection -> IO (Map ChainId Int64)
+countTxMissingEvents pool = do
+    chainCounts <- P.withResource pool $ \c -> runBeamPostgresDebug putStrLn c $
```
mightybyte, line 179: Also these debugs

exec/Chainweb/Env.hs
```diff
@@ -158,11 +159,19 @@ readNodeDbPath = eitherReader $ \case
...
     | Gaps (Maybe Int)
     | Single ChainId BlockHeight
     deriving (Show)
 
+data BackfillArgs = BackfillArgs
+  { _backfillArgs_delayMicros :: Maybe Int
+  , _backfillArgs_onlyEvents :: Bool
+  , _backfillArgs_eventChunkSize :: Maybe Integer
+  , _backfillArgs_chainwebVersion :: ChainwebVersion
+  , _backfillArgs_coinbaseDelayMicros :: Maybe Int
```
mightybyte, line 172: Let's consolidate these two delay args.  Having them separate does give the user more flexibility, but I think in practice it will be more trouble than the user wants to deal with.

exec/Chainweb/Env.hs
```diff
@@ -158,11 +159,18 @@ readNodeDbPath = eitherReader $ \case
...
+    | Backfill BackfillArgs
     | Gaps (Maybe Int)
     | Single ChainId BlockHeight
     deriving (Show)
 
+data BackfillArgs = BackfillArgs
+  { _backfillArgs_delayMicros :: Maybe Int
+  , _backfillArgs_onlyEvents :: Bool
+  , _backfillArgs_eventChunkSize :: Maybe Integer
+  , _backfillArgs_chainwebVersion :: ChainwebVersion
```
mightybyte, line 171: We should remove this field.  This information already exists in the `NodeInfo` data structure from chainweb-api (https://github.com/kadena-io/chainweb-api/blob/master/lib/Chainweb/Api/NodeInfo.hs#L19).

mightybyte (changes requested):

mightybyte (approved):

PR #55, opened 2021-07-20
Separate event filling

mightybyte:
The existing `gaps` command does not fill gaps in the event table because it searches for gaps in the block table.  This PR creates a new standalone command for filling events that looks for gaps in the events table itself.


emmanueldenloye (commented):
LGTM

PR #58, opened 2021-07-30
First sketch of signers table

mightybyte:


exec/Chainweb/Lookups.hs
```diff
@@ -205,6 +209,20 @@ mkBlockEvents height cid blockhash pl =  cbes ++ concatMap snd txes
   where
     (cbes, txes) = mkBlockEvents' height cid blockhash pl
 
+mkTransactionSigners :: CW.Transaction -> [Signer]
+mkTransactionSigners t = zipWith3 mkSigner signers sigs [0..]
+  where
+    signers = _pactCommand_signers $ CW._transaction_cmd t
+    sigs = CW._transaction_sigs t
+    mkSigner signer sig idx = Signer
+      (DbHash $ hashB64U $ CW._transaction_hash t)
```
sirlensalot, line 218: Confirming this is the correct unpadded value?

exec/Chainweb/Lookups.hs
```diff
@@ -219,15 +237,16 @@ bpwoMinerKeys = _minerData_publicKeys . _blockPayloadWithOutputs_minerData
 
 mkTransaction :: Block -> (CW.Transaction, TransactionOutput) -> Transaction
 mkTransaction b (tx,txo) = Transaction
-  { _tx_chainId = _block_chainId b
+  { _tx_requestKey = DbHash $ hashB64U $ CW._transaction_hash tx
```
sirlensalot, line 240: Is this unpadded?

lib/ChainwebDb/Types/Signer.hs
```diff
@@ -0,0 +1,78 @@
...
+import           Database.Beam.Backend.SQL.Row (FromBackendRow)
+import           Database.Beam.Backend.SQL.SQL92 (HasSqlValueSyntax)
+import           Database.Beam.Migrate
+import           Database.Beam.Postgres
+import           Database.Beam.Postgres.Syntax (PgValueSyntax)
+------------------------------------------------------------------------------
+import           ChainwebDb.Types.DbHash
+------------------------------------------------------------------------------
+
+newtype HexText = HexText { unHexText :: Text }
```
sirlensalot, line 34: Hmm, how does this differentiate from our keys which are generally hex? We actually can't assume what format hashes will be in for new schemes. I would suggest a `Signature` or similar to be more specific.

sirlensalot (approved):
Some renaming suggestions mainly, but nothing incorrect. One last question are we happy with `TransactionT` keying on block? Request key captures the entire tx so maybe we should just key on request key, and that way we won't have orphan issues. Dunno if that's in scope for this PR though ...

exec/Chainweb/Lookups.hs
mightybyte: Yes.  I fixed that upstream so we wouldn't have to go around and change all the call sites. https://github.com/kadena-io/chainweb-api/blob/master/lib/Chainweb/Api/Hash.hs#L27

exec/Chainweb/Lookups.hs
mightybyte: Yes, see above.

lib/ChainwebDb/Types/Signer.hs
mightybyte: Oh, that's a good point.  I hadn't thought about other signature schemes.  At the moment public keys keys are just plain `Text`, which I think makes sense because of the presence of the `scheme` field.  I was thinking about this type being generic and usable for any hex-encoded text but it probably does make sense to limit this to just signatures, which I think we can plan on always treating as hex.  I'll rename the data type.

mightybyte:
@sirlensalot I don't believe request key does captures everything because transactions can be reintroduced.  @emmanueldenloye and I ran into this issue when we were debugging the DB filling code.

lib/ChainwebDb/Types/Signer.hs
sirlensalot: We can't assume anything, even hex.

sirlensalot:
I guess since our `Transaction` is actually an _executed_ transaction, that's correct, since the outputs depend on the block.

sirlensalot:
Truly denormalized would be:
- transaction INPUT keyed by just req key
- transaction OUTPUT keyed by req key + block hash

lib/ChainwebDb/Types/Signer.hs
mightybyte: Got it.  I checked and we're actually not assuming hex anywhere here.

mightybyte:
Ahh yeah.  I think this is a good compromise for now because it avoids a join when you want to query on tx success/failure, etc.

PR #61, opened 2021-08-09
Fill command with periodic execution

mightybyte:
Update README
Add fill command that is a duplicate of gaps
Deprecate gaps and backfill
Add an option for server to run fill daily

### DIFF
--- a/exec/Chainweb/Env.hs
+++ b/exec/Chainweb/Env.hs
@@ -18,7 +18,7 @@ module Chainweb.Env
   , ChainwebVersion(..)
   , Command(..)
   , BackfillArgs(..)
-  , GapArgs(..)
+  , FillArgs(..)
   , envP
   , richListP
   , NodeDbPath(..)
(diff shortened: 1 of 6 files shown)


exec/Chainweb/Env.hs
```diff
@@ -258,8 +261,8 @@ bfArgsP = BackfillArgs
   <$> delayP
   <*> optional (option auto (long "chunk-size" <> metavar "CHUNK_SIZE" <> help "Number of transactions to query at a time"))
 
-gapArgsP :: Parser GapArgs
-gapArgsP = GapArgs
+fillArgsP :: Parser FillArgs
+fillArgsP = FillArgs
   <$> delayP
   <*> flag False True (long "disable-indexes" <> short 'd' <> help "Disable indexes on tables while filling on gaps.")
```
emmanueldenloye, line 267: Maybe update the comment here as well (maybe not use the words gaps).

emmanueldenloye (commented):
LGTM (sans my one comment)

exec/Chainweb/Env.hs
mightybyte: Ahh, good catch

PR #62, opened 2021-08-13
Retry when server node listener fails

mightybyte:


exec/Chainweb/Server.hs
```diff
@@ -164,6 +166,19 @@ apiServerCut env senv cutBS = do
   Network.Wai.Handler.Warp.run (_serverEnv_port senv) $ setCors $ \req f ->
     serve theApi (serverApp req) req f
 
+retryingListener :: Env -> IORef ServerState -> IO ()
+retryingListener env ssRef = do
+  let logg = _env_logger env
+      delay = 10_000_000
+      policy = constantDelay delay
```
larskuhtz, line 173: Does this retry forever? Should it fail after some time? Something that would be longer than an expected downtime for a node, say, maybe 10min or so?

I guess, that if it fails the application would exit and the admin would be notified. If it never fails, how would the admin become aware of the issue?

mightybyte: I thought about that.  I think it should retry forever and that is what this does.  I think anyone running an important deployment version of the server probably wants it to stay up to continue serving queries.  If the node stays down it will start spitting out errors and warnings and that should be sufficient to trigger any kind of reaction that might be desired while still continuing to be operational as a server.

larskuhtz (approved):

PR #67, opened 2021-12-07
First pass - optimizing code search and event queries

emmanueldenloye:


exec/Main.hs
```diff
@@ -89,6 +95,21 @@ main = do
       Args _ _ _ _ level _ -> level
       RichListArgs _ level -> level
 
+addTransactionsHeightIndex :: LogFunctionIO Text -> Connection -> IO ()
```
mightybyte, line 98: We should probably make these functions check to see whether the index exists and skip the `create` if it does.

mightybyte (changes requested):
The cabal.project file got deleted.  We need to revert that.

lib/ChainwebDb/Types/Transaction.hs
```diff
@@ -88,6 +92,9 @@ Transaction
 type Transaction = TransactionT Identity
 type TransactionId = PrimaryKey TransactionT Identity
 
+instance HasColumnType TsVector where
```
mightybyte, line 95: I think this might be no longer needed, right?

lib/ChainwebDb/Types/Transaction.hs
```diff
@@ -7,6 +7,7 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeApplications #-}
```
mightybyte, line 10: If the below question is "yes", then this is probably unnecessary as well.

mightybyte (changes requested):

lib/ChainwebDb/Types/Transaction.hs
emmanueldenloye: This can be removed.

mightybyte (approved):

PR #76, opened 2022-03-07
Adding accounts transfer history query endpoint v2

emmanueldenloye:


lib/ChainwebDb/Types/Transfer.hs
```diff
@@ -0,0 +1,62 @@
...
+------------------------------------------------------------------------------
+data TransferT f = Transfer
+  {
+    _tr_creationtime :: C f UTCTime
+  , _tr_block :: PrimaryKey BlockT f
+  , _tr_requestkey :: C f ReqKeyOrCoinbase
+  , _tr_chainid :: C f Int64
+  , _tr_height :: C f Int64
+  , _tr_idx :: C f Int64
+  , _tr_name :: C f Text
```
mightybyte, line 34: I think we should change this to `modulename` or something a little clearer.

exec/Chainweb/FillTransfers.hs
```diff
@@ -0,0 +1,102 @@
...
+    guard_ $ pgJsonTypeOf (_ev_params ev -># val_ 0) ==. (val_ "string")
+    guard_ $ pgJsonTypeOf (_ev_params ev -># val_ 1) ==. (val_ "string")
+    guard_ $
+      pgJsonTypeOf (_ev_params ev -># val_ 2) ==. (val_ "number")
+      ||. pgJsonTypeOf (_ev_params ev -># val_ 2 ->$ "decimal") ==. (val_ "number")
+      ||. pgJsonTypeOf (_ev_params ev -># val_ 2 ->$ "int") ==. (val_ "number")
+    let from_acct = _ev_params ev -># val_ 0
+        to_acct = _ev_params ev -># val_ 1
+        -- this is ugly, but we have to do things the sql way
+        amount = ifThenElse_ (pgJsonTypeOf (_ev_params ev -># val_ 2 ->$ "decimal") ==. val_ "number")
```
mightybyte, line 87: Slick!

mightybyte (changes requested):

cabal.project
```diff
@@ -16,7 +16,7 @@ source-repository-package
 source-repository-package
     type: git
     location: https://github.com/kadena-io/chainweb-api.git
-    tag: 4d3d0ac5108254ac27408101903ee503702873ba
+    tag: a18389bb43f480238aee44398239e1b311adee41
```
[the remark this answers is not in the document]
emmanueldenloye: No, I'll make one.

lib/ChainwebDb/Types/Transfer.hs
```diff
@@ -0,0 +1,66 @@
...
+------------------------------------------------------------------------------
+data TransferT f = Transfer
+  {
+    _tr_creationtime :: C f UTCTime
+  , _tr_block :: PrimaryKey BlockT f
+  , _tr_requestkey :: C f ReqKeyOrCoinbase
+  , _tr_chainid :: C f Int64
+  , _tr_height :: C f Int64
+  , _tr_idx :: C f Int64
+  , _tr_qualName :: C f Text
```
mightybyte, line 34: We should be able to ditch qualName from this table since we're filtering to TRANSFER and we already have modulename.

emmanueldenloye: Yeah that makes sense.

lib/ChainwebDb/Types/Common.hs
```diff
@@ -0,0 +1,61 @@
...
+  show RKCB_Coinbase = "cb"
+  show (RKCB_RequestKey rk) = T.unpack $ unDbHash rk
+
+getTxHash :: ReqKeyOrCoinbase -> T.Text
+getTxHash = \case
+      RKCB_RequestKey txhash -> unDbHash txhash
+      RKCB_Coinbase -> "<coinbase>"
+
+rkcbFromText :: Text -> ReqKeyOrCoinbase
+rkcbFromText "cb" = RKCB_Coinbase
```
mightybyte, line 45: This may need to be `<coinbase>` to match previous behavior.

exec/Chainweb/BackfillTransfers.hs
```diff
@@ -0,0 +1,330 @@
...
+        exitFailure
+      Nothing -> die "IMPOSSIBLE: This query (SELECT EXISTS (SELECT 1 as events);) failed somehow."
+
+    let startingHeight :: Int64 = case _nodeInfo_chainwebVer $ _env_nodeInfo env of
+          "mainnet01" -> 1_722_501
+          "testnet04" -> 1_261_001
+          _ -> error "Chainweb version: Unknown"
+
+    minHeights <- withDbDebug env Debug chainMinHeights
+    let checkMinHeights =
```
mightybyte, line 73: This a little hard to follow.  Might be worth simplification.

exec/Chainweb/BackfillTransfers.hs
```diff
@@ -0,0 +1,330 @@
...
+          maybe False (== _nodeInfo_numChains (_env_nodeInfo env)) . foldr go (Just 0)
+        go (_,mh) (Just acc) =
+          case mh of
+            Just _ ->  Just $ acc + 1
+            Nothing -> Nothing
+        go _ Nothing = Nothing
+    unless (checkMinHeights minHeights) $ do
+      logg Error "Make sure transfers table has an entry for every chain id!"
+      exitFailure
+    let minMinHeights = minimum $ mapMaybe snd $ minHeights
```
mightybyte, line 83: I think this should be `maximum` instead of `minimum`.

exec/Chainweb/BackfillTransfers.hs
```diff
@@ -0,0 +1,330 @@
...
+    return $ _tx_creationTime t
+  where
+    getBlockHashes  = mapMaybe $ \case
+      (bhash, RKCB_Coinbase) -> Just bhash
+      _ -> Nothing
+
+eventSelector :: Int64 -> Int64 -> Maybe Int64 -> Q Postgres ChainwebDataDb s (EventT (QExpr Postgres s))
+eventSelector startingHeight endingHeight chainid = do
+      ev <- all_ (_cddb_events database)
+      guard_ $ _ev_height ev <. val_ endingHeight
```
mightybyte, line 314: I think this should be `<=.` based on the reasoning that we may not have gotten all the entries at `endingHeight` on the previous pass and having a small number of multiple inserts shouldn't be a problem.

enobayram (changes requested):
I think this is a great PR and I'm looking forward to see it merged! 

I've asked for some minor changes in my review comments as well as a few stylistic suggestions.

Here's a list of manual verification steps we could run and document under this PR that could give us a lot of confidence for the correctness of this PR:

#### Verify the equivalence of different way of populating the `transfers` table
1) Run `backfill` on a brand new `chainweb-data` instance and let the `transfers` table get filled as part of the regular backfill.
2) Go to the database and rename the `transfers` table to `transfers_backup`.
3) Run `backfill-transfers` and let `chainweb-data` repopulate the `transfers` table.
4) Go to the database and verify that the `transfers` table is identical to the `transfers_backup` table.
   * NOTE: This should verify that backfill is consistent with backfill-transfers
5) Let the `chainweb-data` instance run in `listen` mode for a while so that it appends some events and transfers, and then stop the `listen`ing process.
6) Go to the database and rename the `transfers` table to `transfers_backup2`.
7) Similar to step 3, run `backfill-transfers` and let `chainweb-data` repopulate the `transfers` table.
   * NOTE: This should verify that backfill@time0 + append@time0-time1 = backfill-transfers@time1

#### Verify that the `transfers` contents make sense 
8) Start the `chainweb-data` server mode
9) Run a `GET /account` query on an account that's expected to return some results
10) Go to the database and fetch the events related to that account and manually confirm that the HTTP response is correct.
   * NOTE: This verifies that the `GET /account` endpoint works as expected
11) Make [a pact `POST /local`](https://api.chainweb.com/openapi/pact.html#tag/endpoint-local) request (or follow some other method that queries the chainweb pact state) to verify that the account balance agrees with the sum of the inbound and outbound transfers

exec/Chainweb/Server.hs
```diff
@@ -416,13 +421,26 @@ txsHandler logger pool (Just (RequestKey rk)) =
...
+  -> Int -- ^ chain identifier
   -> Maybe Limit
   -> Maybe Offset
   -> Handler [AccountDetail]
-accountHandler _logger _pool _token _accountName _chain _limit _offset = throw404 "accounts endpoint has yet to be implemented"
+accountHandler logger pool token account chain limit offset =
+  liftIO $ P.withResource pool $ \c -> do
+    r <- runBeamPostgresDebug (logger Debug . T.pack) c $ runSelectReturningList $ accountQueryStmt limit offset token account chain
+    return $ (`map` r) $ \tr -> AccountDetail
+      { _acDetail_name = _tr_modulename tr
```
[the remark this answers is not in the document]
emmanueldenloye: This is because I'm pointing at another commit of `chainweb-api` that doesn't include this field.

exec/Chainweb/FillTransfers.hs
```diff
@@ -0,0 +1,103 @@
+{-# LANGUAGE DerivingStrategies #-}
```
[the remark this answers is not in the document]
emmanueldenloye: This module has been deleted.

exec/Chainweb/Listen.hs
```diff
@@ -66,18 +68,23 @@ getOutputsAndInsert env ph@(PowHeader h _) = do
...
       !t = mkBlockTransactions b pl
       !es = mkBlockEvents (fromIntegral $ _blockHeader_height $ _hwp_header ph) (_blockHeader_chainId $ _hwp_header ph) (DbHash $ hashB64U $ _blockHeader_hash $ _hwp_header ph) pl
       !ss = concat $ map (mkTransactionSigners . fst) (_blockPayloadWithOutputs_transactionsWithOutputs pl)
+
       !k = bpwoMinerKeys pl
-  writes pool b k t es ss
+  case eventsMinHeight version of
+    Nothing -> die $ printf "insertNewHeader failed because we don't know how to work this version %s" version
+    Just minHeight -> do
+      let !tf = mkTransferRows (fromIntegral $ _blockHeader_height $ _hwp_header ph) (_blockHeader_chainId $ _hwp_header ph) (DbHash $ hashB64U $ _blockHeader_hash $ _hwp_header ph) (posixSecondsToUTCTime $ _blockHeader_creationTime $ _hwp_header ph) pl minHeight
```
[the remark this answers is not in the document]
emmanueldenloye: I've addressed this in a commit. Thanks for the catch!

exec/Main.hs
```diff
@@ -148,6 +152,24 @@ addEventsModuleNameIndex =
     , statement = "CREATE INDEX IF NOT EXISTS events_height_chainid_module ON events (height DESC, chainid, module);"
     }
 
+addFromAccountsIndex :: LogFunctionIO Text -> Connection -> IO ()
+addFromAccountsIndex =
+  addIndex
+    IndexCreationInfo
+    {
+      message = "Adding \"(from_acct)\" index on transfers table"
+    , statement = "CREATE INDEX IF NOT EXISTS ON transfers (from_acct);"
```
enobayram, line 161: There's a SQL syntax error here, `IF NOT EXISTS` need to be followed by the name of the index. The same goes for `addToAccountsIndex` as well.

exec/Chainweb/Server.hs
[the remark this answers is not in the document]
enobayram: Ah, that's interesting. As far as I can tell, [the `chainweb-api` version pointed at](https://github.com/kadena-io/chainweb-data/blob/ff2cf058fde941863cb7ad6541195a5052c12bac/cabal.project#L16-L20) by the cw-data's cabal.project file [contains the `_acDetail_creationTime` field](https://github.com/kadena-io/chainweb-api/blob/4e8e820d4888a5a29087b97798f0d6551593ff33/lib/ChainwebData/AccountDetail.hs#L12), so which `chainweb-api` version are you and the CI pointing at? What file pins to that version, is it [this `nix-thunk` file](https://github.com/kadena-io/chainweb-data/blob/ff2cf058fde941863cb7ad6541195a5052c12bac/deps/chainweb-api/github.json#L1-L9)?

If that's the case, seems to me like we need to unify the pinned `chainweb-api` revisions in `cabal.project` and nix-thunk

lib/ChainwebDb/Queries.hs
```diff
@@ -107,3 +108,28 @@ eventsQueryStmt limit offset qSearch qParam qName qModuleName bh =
...
+    :: Maybe Limit
+    -> Maybe Offset
+    -> Text
+    -> Text
+    -> Int
+    -> SqlSelect
+    Postgres
+    (QExprToIdentity
+    (TransferT (QGenExpr QValueContext Postgres QBaseScope)))
+accountQueryStmt limit offset token account chain =
```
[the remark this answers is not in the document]
emmanueldenloye: I think we've covered this in another discussion.

exec/Main.hs
emmanueldenloye: I've resolved this issue.

exec/Chainweb/Lookups.hs
```diff
@@ -204,11 +216,88 @@ mkBlockEvents' height cid blockhash pl =
...
+          , _tr_to_acct =
+              case unwrap (_ev_params ev) ^? ix 1 of
+                Just (String s) -> s
+                _ -> error "mkTransferRows: to_account is not a string"
+          , _tr_amount = either error id $ do
+                let msg = "mkTransferRows: amount is somehow missing"
+                param <- note msg $ unwrap (_ev_params ev) ^? ix 2
+                parseEither (\v -> (fromRational @Double . toRational <$> decoder decimalCodec v) <|> (fromInteger <$> decoder integerCodec v)) param
+          }
+    createNonCoinBaseTransfers xs =
```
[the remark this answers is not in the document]
emmanueldenloye: I've addressed this.

lib/ChainwebDb/Types/Transfer.hs
```diff
@@ -0,0 +1,60 @@
...
+  (LensFor tr_requestkey)
+  (LensFor tr_chainid)
+  (LensFor tr_height)
+  (LensFor tr_idx)
+  (LensFor tr_modulename)
+  (LensFor tr_moduleHash)
+  (LensFor tr_from_acct)
+  (LensFor tr_to_acct)
+  (LensFor tr_amount)
+ = tableLenses
```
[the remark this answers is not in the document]
emmanueldenloye: It has been the standard for the other tables to include these lenses. I don't necessarily see the harm in keeping them around. But I'll let you decide what's best here since I'm not really leaning one way or the other.

exec/Chainweb/BackfillTransfers.hs
```diff
@@ -0,0 +1,138 @@
...
+      <*> pure (_ev_module ev)
+      <*> pure (_ev_moduleHash ev)
+      <*> from_acct
+      <*> to_acct
+      <*> amount
+  where
+    amount = case _ev_params ev ^? to unwrap . ix 2 . key "decimal" of
+      Just (A.Number n) -> Just $ toRealFloat n
+      _ -> case _ev_params ev ^? to unwrap . ix 2 . key "int" of
+        Just (A.Number n) -> Just $ toRealFloat n
```
mightybyte, line 118: Apparently Postgres won't let you cast from jsonb to decimal (see https://www.postgresql.org/message-id/0154d35a-24ae-f063-5273-9ffcdf1c7f2e%40postgrespro.ru).  I'd like to verify that the SQL expressions we're using to convert data in the events table into the transfers table is handling this correctly.  Can we do a search for transactions that have an amount in the format `{"decimal": "1.234"}` and verify that those rows are indeed showing up in the transfers table?

lib/ChainwebDb/Types/Transfer.hs
```diff
@@ -0,0 +1,104 @@
...
+https://www.postgresql.org/docs/current/datatype-numeric.html.
+
+Since the maximum amount of kda someone can have in their posssession is around
+1 billion and also the maximum precision of any number represented by pact needs
+12 digits past the decimal point, we have chosen the numbers 21 and 12 for the
+precision and scale of the newtype KDAScientific.
+
+-}
+
+newtype KDAScientific = KDAScientific { getKDAScientific :: Scientific }
```
enobayram, line 84: Notes from today's call: We need to make sure that this representation is the same as the one we have in chainweb-api. In particular we're changing the number type to `Decimal` here: https://github.com/kadena-io/chainweb-api/pull/29

We also need to change the `beam-migrate` interpretation of this type to use the `numeric` Postgres field type. We had a discussion about this earlier and the result was that Postgres's `numeric` is very flexible and it doesn't take up more space than needed. Observe:
``` SQL
=> select pg_column_size(a)
from (values
    (1::numeric),
    (1e100::numeric),
    (1e-100::numeric),
    (1e100-1e-100::numeric) 
) s(a)       
;
 pg_column_size 
----------------
              8
              8
             10
            108
(4 rows)
```

enobayram:
Here's what remains here:
- [ ] Address this comment: https://github.com/kadena-io/chainweb-data/pull/76/files#r964135523
- [ ] Pin chainweb-api to the latest commit on master (both for nix and in `cabal.project)
- [ ] Perform the manual test steps here: https://github.com/kadena-io/chainweb-data/pull/76#pullrequestreview-1066592686

emmanueldenloye:
Here is a script to verify that the transfers table is filled correctly under all circumstances https://gist.github.com/emmanueldenloye/b814a07531afd06b9c6547bc66c5d902

Here is some command-line output showing that the accounts endpoint produces correct output https://gist.github.com/emmanueldenloye/cc7f6b14fa92606585f41ae71af45d46

lib/ChainwebDb/Types/Common.hs
```diff
@@ -0,0 +1,61 @@
...
+  deriving (Eq, Ord, Generic)
+
+instance Show ReqKeyOrCoinbase where
+  show RKCB_Coinbase = "cb"
+  show (RKCB_RequestKey rk) = T.unpack $ unDbHash rk
+
+getTxHash :: ReqKeyOrCoinbase -> T.Text
+getTxHash = \case
+      RKCB_RequestKey txhash -> unDbHash txhash
+      RKCB_Coinbase -> "<coinbase>"
```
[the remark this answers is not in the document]
emmanueldenloye: This string appears in the events endpoint as well. We shouldn't make a backwards incompatible change in this PR.

enobayram (approved):
LGTM! Great work on this PR, glad it's finally getting merged.

mightybyte (approved):

PR #90, opened 2022-08-15
Improve server endpoints

emmanueldenloye:


enobayram (approved):
LGTM! Thanks for pulling these queries to the top-level, should make it much easier to play with them.

PR #94, opened 2022-09-20
Change numeric type of transfer amount type to Decimal

emmanueldenloye:
We concluded that Decimal is the best way to go given these considerations:

1. We could store the  amount as json, especially considering that the corresponding events sometimes stores the amount as json. 

Pro(s) - We don't have to think very hard about how to represent the data and their is no potential loss of information.
Con(s) - We ultimately push the problem of parsing the number correctly to the consumer.

2. We could simply write the number as a string to postgres.

Pro(s) - This is easy to implement and parse. Easy to reason over what gets stored in the database and passed back through over the Haskell interface.
Con(s) - This is an inefficient use of space.

3. We could write the number as a type that could be translated to a numeric type in postgres.

Pro(s) - "natural" representation for later analysis AND potentially most efficient use of storage.

Con(s) - Could be hard to get an accurate representation.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -16,7 +16,7 @@ source-repository-package
 source-repository-package
     type: git
     location: https://github.com/kadena-io/chainweb-api.git
-    tag: 4e8e820d4888a5a29087b97798f0d6551593ff33
+    tag: b7fdbe35dbe4e7b818dc9650fdb950ea5eb7990f
 
 source-repository-package
     type: git
--- a/deps/chainweb-api/github.json
+++ b/deps/chainweb-api/github.json
@@ -3,6 +3,6 @@
   "repo": "chainweb-api",
   "branch": "master",
   "private": false,
-  "rev": "27d45461841ec0c8527c3b8854b2eeda405607a0",
+  "rev": "b7fdbe35dbe4e7b818dc9650fdb950ea5eb7990f",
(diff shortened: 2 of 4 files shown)


enobayram:
Here's how we can manually verify that `KDAScientific` works well with `beam`:
1) Create a Haskell `data` type that has a `KDAScientific` column.
2) Let `beam` create a table for that `data` type.
3) `psql` into your Postgres database and query the table schema `\d mytable`
4) In `ghci` wrtie a `KDAScientific 1e1000` row to that table
5) Read it back and confirm that the value is `1e1000`.

emmanueldenloye:
> Here's how we can manually verify that `KDAScientific` works well with `beam`:
> 
> 1. Create a Haskell `data` type that has a `KDAScientific` column.
> 2. Let `beam` create a table for that `data` type.
> 3. `psql` into your Postgres database and query the table schema `\d mytable`
> 4. In `ghci` wrtie a `KDAScientific 1e1000` row to that table
> 5. Read it back and confirm that the value is `1e1000`.

Here are the steps necessary to replicate what is above. https://gist.github.com/emmanueldenloye/8e3c54d1419bb19aa471abebb2a0e01e

enobayram (approved):
LGTM!

Issue #79, opened 2022-06-15
Missing information about indexes required for indexer's endpoints


blazesal:
#### Naming
In the description below following names are being used for:
- Postgres super-user: `postgres`,
- database name: `chainweb-data`,
- database user: `db_user`.

#### Increase debug query size
Increase the query size, so it is possible to display large queries:
```
sudo -u postgres psql chainweb-data
ALTER SYSTEM SET track_activity_query_size = 16384;
quit
systemctl restart postgresql
```

#### Intercept long-running query
Invoke the indexer's endpoint, which fails to return the data quickly:
```
curl http://localhost:8080/txs/search?limit=200&search=k:5adb16663073280acf63bc2a4bf477ad1391736dcd6217b094926862c72d15c9
```
Connect to the database and display the running SQL queries:
```
sudo -u db_user psql -d chainweb-data -U db_user -W
SELECT pid, age(clock_timestamp(), query_start), usename, query FROM pg_stat_activity WHERE query != '<IDLE>' AND query NOT ILIKE '%pg_stat_activity%' ORDER BY query_start desc;
```
Then analyze the query:
```
explain analyze SELECT "t0"."chainid" AS "res0", "t1"."height" AS "res1", "t0"."block" AS "res2", "t0"."creationtime" AS "res3", "t0"."requestkey" AS "res4", "t0"."sender" AS "res5", "t0"."code" AS "res6", "t0"."continuation" AS "res7", "t0"."goodresult" AS "res8" FROM "transactions" AS "t0" CROSS JOIN "blocks" AS "t1" WHERE (("t0"."block") = ("t1"."hash")) AND (("t0"."code") LIKE ('%k:5adb16663073280acf63bc2a4bf477ad1391736dcd6217b094926862c72d15c9%')) ORDER BY "t1"."height" DESC LIMIT 100 OFFSET 0
```
Analysis gives the following query plan:
```
"QUERY PLAN"
"Limit  (cost=556941.22..556952.89 rows=100 width=1064) (actual time=2328.803..2336.257 rows=81 loops=1)"
"  ->  Gather Merge  (cost=556941.22..556990.92 rows=426 width=1064) (actual time=2105.215..2112.657 rows=81 loops=1)"
"        Workers Planned: 2"
"        Workers Launched: 2"
"        ->  Sort  (cost=555941.20..555941.73 rows=213 width=1064) (actual time=2093.093..2093.099 rows=27 loops=3)"
[...]

PR #98, opened 2022-11-29
Add requestkey indexes on events and transactions

enobayram:
As identified by @blazesal in https://github.com/kadena-io/chainweb-data/issues/79, our `/txs/tx` endpoint requires a requestkey index on the events and transactions tables, otherwise each request results in a table scan.

I've just tried creating these indexes manually and checking their sizes; As of now, the requestkey index on the transactions table takes up 603MB of space while the one on events takes up 879MB. For comparison, these total space occupied by these two tables is ~46GB.


emmanueldenloye (commented):
LGTM

emmanueldenloye (approved):

PR #111, opened 2022-12-22
Add a CLI option to ignore detected schema diffs

enobayram:
This PR adds an `--ignore-schema-diff` CLI option to the `chainweb-data` binary that makes it keep going even if it encounters a different schema in the database.

This option is useful for `chainweb-data` operators that would like to make ad-hoc additions to their CW-D database. Though it should probably be mentioned that making such ad-hoc changes, especially directly on the tables populated and queried by `chainweb-data` isn't officially supported since an arbitrarily different schema might cause CW-D to break in unpredictable ways.

### DIFF
--- a/lib/ChainwebData/Env.hs
+++ b/lib/ChainwebData/Env.hs
@@ -58,7 +58,7 @@ import           Text.Printf
 
 ---
 
-data MigrateStatus = RunMigration | DontMigrate
+data MigrateStatus = RunMigration | DontMigrate Bool
   deriving (Eq,Ord,Show,Read)
 
 data Args
@@ -215,8 +215,13 @@ envP = Args
   <*> migrationP
 
 migrationP :: Parser MigrateStatus
-migrationP =
-  flag DontMigrate RunMigration (short 'm' <> long "migrate" <> help "Run DB migration")
+migrationP
+    = flag' RunMigration (short 'm' <> long "migrate" <> help "Run DB migration")
+  <|> flag' (DontMigrate False)
+        ( long "--ignore-schema-diff"
+       <> help "Ignore any unexpected differences in the database schema"
+        )
+  <|> pure (DontMigrate True)
 
 logLevelParser :: Parser LogLevel
 logLevelParser =
--- a/lib/ChainwebData/Env.hs
+++ b/lib/ChainwebData/Env.hs
@@ -217,11 +217,11 @@ envP = Args
 migrationP :: Parser MigrateStatus
 migrationP
     = flag' RunMigration (short 'm' <> long "migrate" <> help "Run DB migration")
-  <|> flag' (DontMigrate False)
-        ( long "--ignore-schema-diff"
+  <|> flag' (DontMigrate True)
+        ( long "ignore-schema-diff"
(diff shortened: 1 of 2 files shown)


lib/ChainwebData/Env.hs
```diff
@@ -215,8 +215,13 @@ envP = Args
   <*> migrationP
 
 migrationP :: Parser MigrateStatus
-migrationP =
-  flag DontMigrate RunMigration (short 'm' <> long "migrate" <> help "Run DB migration")
+migrationP
+    = flag' RunMigration (short 'm' <> long "migrate" <> help "Run DB migration")
+  <|> flag' (DontMigrate False)
+        ( long "--ignore-schema-diff"
```
alber70g, line 221:
```suggestion
        ( long "ignore-schema-diff"
```

lib/ChainwebData/Env.hs
```diff
@@ -215,8 +215,13 @@ envP = Args
...
+       <> help "Ignore any unexpected differences in the database schema"
+        )
+  <|> pure (DontMigrate True)
```
alber70g, line 224:
```suggestion
  <|> flag' (DontMigrate True)
        ( long "--ignore-schema-diff"
       <> help "Ignore any unexpected differences in the database schema"
        )
  <|> pure (DontMigrate False)
```

enobayram:
Here's how I tested this PR:

I went to my CW-D database and created a `tmp` table:
``` SQL
CREATE TABLE tmp (block TEXT NOT NULL REFERENCES blocks(hash));
```

With `tmp` in the database, trying to run `chainweb-data` from the current master fails with:
```
/nix/store/gfly46vsbkxjvmf6c42nfc80b6x91irk-chainweb-data-2.0.1/bin/chainweb-data --dbstring ... --service-host localhost --p2p-host localhost --p2p-port ... fill
2022-12-23T12:22:37.232Z [Info] [] Database needs to be migrated.  Re-run with the -m option or you can migrate by hand with the following query:
        ALTER TABLE tmp DROP CONSTRAINT tmp_block_fkey;
```
I get the same failure (with a slightly different error message) from this branch when I run it it without `--ignore-schema-diff`:
```
$ cabal exec chainweb-data -- --dbstring ... --service-host localhost --p2p-host localhost --p2p-port ... fill
2022-12-23T12:24:25.221Z [Info] [] Database needs to be migrated.
        ALTER TABLE tmp DROP CONSTRAINT tmp_block_fkey;


2022-12-23T12:24:25.251Z [Error] [] Re-run with the -m option or you can run the queries above manually
```
But it works with `--ignore-schema-diff`:
```
2022-12-23T12:25:26.919Z [Info] [] Database needs to be migrated.
        ALTER TABLE tmp DROP CONSTRAINT tmp_block_fkey;


2022-12-23T12:25:26.948Z [Info] [] Adding height index on transactions table
NOTICE:  relation "transactions_height_idx" already exists, skipping
...
2022-12-23T12:25:26.950Z [Info] [] DB Tables Initialized
...
```

emmanueldenloye (approved):
LGTM

PR #113, opened 2022-12-22
fix: invert values for DontMigrate. Define long flag without dashes

alber70g:
- Add a CLI option to ignore detected schema diffs
- fix: invert values for `DontMigrate`. Define long flag without dashes


enobayram:
Thanks for fixing these issues!

PR #116, opened 2022-12-23
Add load-based throttling to search endpoints

enobayram:
This PR adds instrumentation around the connection pool used by the endpoint handlers to keep track of the number of requests currently pending or being served. It then uses this information in order to scale down the amount of work that's performed per request by the search endpoints in order to make it easier to maintain a fair resource allocation under load across multiple clients.

This PR still doesn't stop a single client from making a large number of search requests in parallel, but `chainweb-data` expects that this would be handled via rate limiting at the API gateway layer. In theory, it could be argued that it would be better for the gateway to handle such throttling as well, but it seems harder to implement there in practice.

This PR also increases the maximum `scanLimit` of the search endpoints from 20000 to 50000 (but subject to load throttling) in order to reduce network roundtrips. It used to take us around 50-100ms to scan 20000 rows, so now each request should take between 100-250ms, which is still low enough and now each request should scan around 10 hours worth of blockchain history.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -16,7 +16,7 @@ source-repository-package
 source-repository-package
     type: git
     location: https://github.com/kadena-io/chainweb-api.git
-    tag: 6ffcf6265309bcd08034652523c940e4b71e9153
+    tag: 953019e5eedc1d23833cb7b5b6c5f53627ff2b20
 
 source-repository-package
     type: git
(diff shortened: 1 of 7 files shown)


emmanueldenloye (approved):
LGTM.... but you should be more careful  next time.

enobayram:
@emmanueldenloye, I've made significant changes to this PR since your approval (see the PR description), so I've asked for another review.

emmanueldenloye (approved):
LGTM

PR #126, opened 2023-01-19
Rename chainid,name to chain,token

enobayram:
That's because:
* We have a query parameter called "token" that's used to filter on that field, so no reason to have a different name, and "token" makes more sense anyway
* The same goes with the "chain" query parameter, but "chainid" is also inconsistent with the events and search endpoints

Note that this is a breaking change on top of v2.1.0, but I think the improvement in consistency is worth it, especially considering that we've released v2.1.0 2 days ago and haven't yet announced it over any of the official or unofficial channels.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -16,7 +16,7 @@ source-repository-package
 source-repository-package
     type: git
     location: https://github.com/kadena-io/chainweb-api.git
-    tag: ca794bec5e7b3a7e58f4862459b712b78191971b
+    tag: 38eb85b5395b575aa2efeff91fbfeb5cb3d494ce
 
 source-repository-package
     type: git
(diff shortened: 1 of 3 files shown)


enobayram:
Here's a simple request demonstrating the change:
```
curl "$CWDURL/txs/account/c50b9acb49ca25f59193b95b4e08e52e2ec89fa1bf308e6436f39a40ac2dc4f3?limit=1" | jq -C
[
  {
    "blockTime": "2023-01-19T13:36:10.101853Z",
    "amount": "1.032494",
    "height": 3390144,
    "blockHash": "1U3iW0qS4JARA7221rYrL02J-L3i_q-K8AyGRQn3IRU",
    "token": "coin",
    "requestKey": "<coinbase>",
    "fromAccount": "",
    "toAccount": "c50b9acb49ca25f59193b95b4e08e52e2ec89fa1bf308e6436f39a40ac2dc4f3",
    "idx": 0,
    "chain": 9
  }
]
```

Takadenoshi (approved):

emmanueldenloye (approved):
LGTM

PR #129, opened 2023-01-20
Bump v2.1.1

enobayram:
Shortly after the 2.1.0 release, we've noticed an oversight in the new `/txs/account` endpoint and decided to correct it with a quick breaking change instead of unnecessarily complicating the API. See [PR #126](https://github.com/kadena-io/chainweb-data/pull/126)


emmanueldenloye (approved):
LGTM

PR #134, opened 2023-02-07
Fix import of cw-data exe for docker

emmanueldenloye:


enobayram:
Thanks for fixing that @emmanueldenloye! As you've noted yesterday, this `nix-build docker.nix` produces a derivation that has a different output hash compared to `nix build .#chainweb-data-docker`, I believe that's because `default.nix` defines its own default for the `pkgs` argument, while `flake.nix` passes its own `pkgs` into `docker.nix` while importing it. I think it could be nice to make `docker.nix` use that same `pkgs` by default. It would be nice to have `docker.nix` and `flake.nix` produce the same output, but another benefit would be avoiding a hard-coded `nixpkgs` dependency in `docker.nix` that's likely to get outdated and go unnoticed.

enobayram (approved):
LGTM!

enobayram (approved):
LGTM Mk.2!

PR #136, opened 2023-02-21
Nix reorganization

enobayram:
This PR turns our Nix configuration somewhat inside out and moves all the heavy lifting to `default.nix` from `flake.nix`.

The rationale is that Nix flakes are inflexible by design so that they provide static information about their inputs and their outputs. By moving the interesting logic back to `default.nix` and importing it from `flake.nix` we're able to accept arguments to the `nix-build` call to build our artifacts with parameters.

This PR introduces the `dontStrip`, `threaded` and `enableProfiling` parameters to `default.nix`.

In order to expose this new flexibility to our `chanweb-data-docker` images, this PR also inlines `docker-legacy.nix` into `default.nix` so that, say, `nix-build --arg enableProfiling true -A dockerImage` builds a docker image with a CW-D binary built with profiling.

Note that this `default.nix` still piggy backs on the new nix flakes feature for managing its inputs like `haskell.nix` and `nixpkgs`.


enobayram:
Here are some steps to verify that the various `default.nix` arguments introduced by this PR work as intended: 

## The default build
(strips the debug symbols, is threaded and has no profiling)
```
$ nix log $(nix-build) | grep -e ".\{0,20\}strip" -e ".\{0,10\}threaded.\{0,10\}" -e ".\{0,10\}library-profiling.\{0,10\}" -o
s --flags=threaded --exact-c
--with-ar=ar --with-strip
=strip
--enable-executable-strip
ng --enable-library-strip
--disable-library-profiling --disable
strip
-wrapper-11.3.0/bin/strip
strip
ping (with command strip
strip
-wrapper-11.3.0/bin/strip
```

The stripped executable size is 16.8MB:
```
$ du -d0 $(nix-build)
16868	/nix/store/zvjl5p532k53bhxis2wipwgw1nf5nyk3-chainweb-data-exe-chainweb-data-2.1.1
```

## --arg dontStrip true

```
$ nix log $(nix-build --arg dontStrip true) | grep -e ".\{0,10\}with-strip.\{0,10\}"  -o
h-ar=ar --with-strip=strip --d
```

Unstripped size is 24.6 MB:
```
$ du -d0 $(nix-build --arg dontStrip true)
24604	/nix/store/yrj4h59cc6zasrysqjxmm4s8b2qqrlax-chainweb-data-exe-chainweb-data-2.1.1
```

## --arg threaded false
````
$ nix log $(nix-build --arg threaded false) | grep -e ".\{0,10\}threaded.\{0,10\}" -o
 --flags=-threaded --exact-c
```
Can't be run threaded:
```
$ $(nix-build --arg threaded false)/bin/chainweb-data +RTS -N
chainweb-data: the flag -N requires the program to be built with -threaded
...
```

## --arg enableProfiling true
```
$ nix log $(nix-build --arg enableProfiling true) | grep -e ".\{0,10\}library-profiling.\{0,10\}" -o
 --enable-library-profiling --enable-
```

emmanueldenloye (approved):

PR #119, opened 2023-01-04
Optimize the recent transactions query

enobayram:
This PR adds instrumentation around the connection pool used by the endpoint handlers to keep track of the number of requests currently pending or being served. It then uses this information in order to scale down the amount of work that's performed per request by the search endpoints in order to make it easier to maintain a fair resource allocation under load across multiple clients.

This PR still doesn't stop a single client from making a large number of search requests in parallel, but `chainweb-data` expects that this would be handled via rate limiting at the API gateway layer. In theory, it could be argued that it would be better for the gateway to handle such throttling as well, but it seems harder to implement there in practice.

This PR also increases the maximum `scanLimit` of the search endpoints from 20000 to 50000 (but subject to load throttling) in order to reduce network roundtrips. It used to take us around 50-100ms to scan 20000 rows, so now each request should take between 100-250ms, which is still low enough and now each request should scan around 10 hours worth of blockchain history.

PR #66, opened 2021-08-18
Search related continuations in tx search

sirlensalot:
This PR extends the transaction endpoints to gather the history of continuation transactions. It extends the following endpoints and adds the `initialCode` and `previousSteps` fields to them:
* Single transaction details: `/txs/tx`
* Multiple transaction details: `/txs/txs`
* Transaction search: `/txs/search`
* Recent transactions: `/txs/recent`

In addition to attaching this continuation history to the response, this PR also extends the search logic of `/txs/search` so that if the given search term appears in the `initialCode` of a continuation transaction, the search still finds it.

This PR supports https://github.com/kadena-io/block-explorer/issues/64 and https://github.com/kadena-io/block-explorer/issues/63

PR #140, opened 2023-03-24
Simplify the handling of the recentTxs

enobayram:
This PR removes the recentTxs tracking from the CW-D server state and always makes a DB query when recent transactions are needed.

With the current indexes that we have in place, the recent txs query is very fast to execute (see #119), so there's no need to complicate the server state for it anymore.

Note that, in addition to simplifying the codebase, this PR is also needed for the continuations search improvements mentioned [here](https://github.com/kadena-io/chainweb-data/pull/66#issuecomment-1453688877) as well as for the planned work of decoupling the node listener from the HTTP API server.

### DIFF
--- a/haskell-src/exec/Chainweb/Server.hs
+++ b/haskell-src/exec/Chainweb/Server.hs
@@ -97,19 +97,10 @@ setCors = cors . const . Just $ simpleCorsResourcePolicy
     }
 
 data ServerState = ServerState
-    { _ssRecentTxs :: RecentTxs
-    , _ssHighestBlockHeight :: BlockHeight
+    { _ssHighestBlockHeight :: BlockHeight
     , _ssTransactionCount :: Maybe Int64
     , _ssCirculatingCoins :: Decimal
     } deriving (Eq,Show)
-
-ssRecentTxs
-    :: Functor f
-    => (RecentTxs -> f RecentTxs)
-    -> ServerState -> f ServerState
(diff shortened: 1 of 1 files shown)


enobayram:
Here's the query that gets executed by the recent transactions handler:
```
=> EXPLAIN ANALYZE 
SELECT "t0"."chainid" AS "res0", "t0"."height" AS "res1", "t0"."block" AS "res2", "t0"."creationtime" AS "res3", "t0"."requestkey" AS "res4", "t0"."sender" AS "res5", "t0"."code" AS "res6", "t0"."continuation" AS "res7", "t0"."goodresult" AS "res8" 
FROM "transactions" AS "t0" 
ORDER BY "t0"."height" DESC 
LIMIT 20;
                                                                                QUERY PLAN                                                                                
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.43..24.48 rows=20 width=1127) (actual time=0.042..0.113 rows=20 loops=1)
   ->  Index Scan Backward using transactions_height_idx on transactions t0  (cost=0.43..16022668.97 rows=13324521 width=1127) (actual time=0.040..0.106 rows=20 loops=1)
 Planning Time: 0.166 ms
 Execution Time: 0.153 ms
(4 rows)
```

So it should be OK to perform it every time its's needed.

emmanueldenloye (approved):

PR #143, opened 2023-04-11
Search transaction by pactid

emmanueldenloye:
A user can now supply either a transaction pact id or a code string (for searching code) to the transaction search endpoint. We've added this functionality so that block-explorer users can click on links allowing them to traverse forward/backwards between the steps of a continuation (pacts).

### DIFF
--- a/haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
+++ b/haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
@@ -1,3 +1,3 @@
 CREATE INDEX transactions_pactid_index 
-  ON transactions (pactid, goodresult DESC, height DESC) 
+  ON transactions (pactid, (goodresult IS NOT NULL), height DESC)
   WHERE pactid IS NOT NULL;
--- a/haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
+++ b/haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
@@ -1,3 +1,3 @@
 CREATE INDEX transactions_pactid_index 
-  ON transactions (pactid, (goodresult IS NOT NULL), height DESC)
+  ON transactions (pactid, (goodresult IS NOT NULL) DESC, height DESC)
   WHERE pactid IS NOT NULL;
--- a/haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
+++ b/haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
@@ -1,0 +1,3 @@
+CREATE INDEX pactIdIndex ON transactions using (pactid) where (pactid is not null);
+-- maybe we could do the following:
+-- CREATE INDEX pactIdIndex ON transactions using (height,pactid) where (pactid is not null);
--- a/haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
+++ b/haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
@@ -1,3 +1,3 @@
-CREATE INDEX pactIdIndex ON transactions using (pactid) where (pactid is not null);
--- maybe we could do the following:
--- CREATE INDEX pactIdIndex ON transactions using (height,pactid) where (pactid is not null);
+CREATE INDEX transactions_pactid_idx
+  ON transactions (pactid)
+  WHERE (pactid IS NOT NULL);
--- a/haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
+++ b/haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
@@ -1,3 +1,3 @@
-CREATE INDEX transactions_pactid_idx
-  ON transactions (pactid)
-  WHERE (pactid IS NOT NULL);
+CREATE INDEX transactions_pactid_index 
+  ON transactions (pactid, goodresult DESC, height DESC) 
+  WHERE pactid IS NOT NULL;
--- a/haskell-src/exec/Chainweb/Server.hs
+++ b/haskell-src/exec/Chainweb/Server.hs
@@ -485,20 +485,18 @@ queryTxsByPactId :: LogFunctionIO Text -> Text -> Connection -> IO [TxSummary]
 queryTxsByPactId logger pactid c =
   runBeamPostgresDebug (logger Debug . T.pack) c $ do
     r <- runSelectReturningList (queryTxsByPactId' pactid)
-    return $ (`fmap` r) $ \(tx,contHist, blk) -> toApiTxSummary tx contHist blk
+    return $ (`fmap` r) $ \(tx,contHist) -> dbToApiTxSummary (toDbTxSummary tx) contHist
 
-queryTxsByPactId' :: Text -> SqlSelect Postgres (TransactionT Identity, ContinuationHistoryT Identity, BlockT Identity)
+queryTxsByPactId' :: Text -> SqlSelect Postgres (TransactionT Identity, ContinuationHistoryT Identity)
 queryTxsByPactId' pactid =
     select $ orderBy_ statusOrd $ do
       tx <- all_ (_cddb_transactions database)
       contHist <- joinContinuationHistory (_tx_pactId tx)
-      blk <- all_ (_cddb_blocks database)
-      guard_ (_tx_block tx `references_` blk)
       let searchExp = val_ (Just $ DbHash pactid)
       guard_' (_tx_pactId tx ==?. searchExp)
-      return (tx,contHist, blk)
+      return (tx,contHist)
   where
-    statusOrd (tx,_,_) = (desc_ (_tx_goodResult tx), desc_ (_tx_height tx))
+    statusOrd (tx,_) = (desc_ (_tx_goodResult tx), desc_ (_tx_height tx))
 
 txHandler
   :: LogFunctionIO Text
--- a/haskell-src/exec/Chainweb/Server.hs
+++ b/haskell-src/exec/Chainweb/Server.hs
@@ -369,14 +369,15 @@ searchTxs
   -> Maybe Limit
   -> Maybe Offset
   -> Maybe Text
+  -> Maybe Text
   -> Maybe BlockHeight -- ^ minimum block height
   -> Maybe BlockHeight -- ^ maximum block height
   -> Maybe NextToken
   -> Handler (NextHeaders [TxSummary])
-searchTxs _ _ _ _ _ Nothing _ _ _ = throw404 "You must specify a search string"
-searchTxs logger pool req givenMbLim mbOffset (Just search) minheight maxheight mbNext = do
+searchTxs _ _ _ _ _ Nothing Nothing _ _ _ = throw404 "You must specify a search string"
+searchTxs logger pool req givenMbLim mbOffset (Just search) Nothing minheight maxheight mbNext = do
   liftIO $ logger Info $ fromString $ printf
-    "Transaction search from %s: %s" (show $ remoteHost req) (T.unpack search)
+    "Transaction search from %s: Search string: %s" (show $ remoteHost req) (T.unpack search)
   continuation <- mkContinuation readTxToken mbOffset mbNext
 
   isBounded <- isBoundedStrategyM req
@@ -398,6 +399,33 @@ searchTxs logger pool req givenMbLim mbOffset (Just search) minheight maxheight 
         resultLimit
       return $ maybe noHeader (addHeader . mkTxToken) mbCont $
         results <&> \(txwh,_) -> dbToApiTxSummary (txwhSummary txwh) (txwhContHistory txwh)
+searchTxs logger pool req givenMbLim mbOffset Nothing (Just pactid) minheight maxheight mbNext = do
+  liftIO $ logger Info $ fromString $ printf
+    "Transaction search from %s: PactId string: %s" (show $ remoteHost req) (T.unpack pactid)
+
+  continuation <- mkContinuation readTxToken mbOffset mbNext
+
+  isBounded <- isBoundedStrategyM req
+
+  liftIO $ M.with pool $ \(c, throttling) -> do
+    let
+      scanLimit = ceiling $ 50000 * throttling
+      maxResultLimit = ceiling $ 250 * throttling
+      resultLimit = min maxResultLimit $ maybe 10 unLimit givenMbLim
+      strategy = if isBounded then Bounded scanLimit else Unbounded
+
+    PG.withTransactionLevel PG.RepeatableRead c $ do
+      (mbCont, results) <- performBoundedScan strategy
+        (runBeamPostgresDebug (logger Debug . T.pack) c)
+        (toTxSearchCursor . txwhSummary)
+        (txSearchSourcePactId pactid $ HeightRangeParams minheight maxheight)
+        noDecoration
+        continuation
+        resultLimit
+      return $ maybe noHeader (addHeader . mkTxToken) mbCont $
+        results <&> \(txwh,_) -> dbToApiTxSummary (txwhSummary txwh) (txwhContHistory txwh)
+searchTxs _logger _pool _req _givenMbLim _mbOffset (Just _search) (Just _pactId) _minheight _maxheight _mbNext =
+  throw404 "You should only specify a pactid or a search string, not both!"
 
 throw404 :: MonadError ServerError m => ByteString -> m a
 throw404 msg = throwError $ err404 { errBody = msg }
--- a/haskell-src/exec/Chainweb/Server.hs
+++ b/haskell-src/exec/Chainweb/Server.hs
@@ -494,9 +494,10 @@ queryTxsByPactId' pactid =
       contHist <- joinContinuationHistory (_tx_pactId tx)
       let searchExp = val_ (Just $ DbHash pactid)
       guard_' (_tx_pactId tx ==?. searchExp)
+      guard_ (isJust_ $ _tx_goodResult tx)
       return (tx,contHist)
   where
-    statusOrd (tx,_) = (desc_ (_tx_goodResult tx), desc_ (_tx_height tx))
+    statusOrd (tx,_) = desc_ (_tx_height tx)
 
 txHandler
   :: LogFunctionIO Text
--- a/haskell-src/exec/Chainweb/Server.hs
+++ b/haskell-src/exec/Chainweb/Server.hs
@@ -494,10 +494,9 @@ queryTxsByPactId' pactid =
       contHist <- joinContinuationHistory (_tx_pactId tx)
       let searchExp = val_ (Just $ DbHash pactid)
       guard_' (_tx_pactId tx ==?. searchExp)
-      guard_ (isJust_ $ _tx_goodResult tx)
       return (tx,contHist)
   where
-    statusOrd (tx,_) = desc_ (_tx_height tx)
+    statusOrd (tx,_) = (desc_ (isJust_ (_tx_goodResult tx)), desc_ (_tx_height tx))
 
 txHandler
   :: LogFunctionIO Text
--- a/haskell-src/exec/Chainweb/Server.hs
+++ b/haskell-src/exec/Chainweb/Server.hs
@@ -399,15 +399,11 @@ searchTxs logger pool req givenMbLim mbOffset (Just search) Nothing minheight ma
         resultLimit
       return $ maybe noHeader (addHeader . mkTxToken) mbCont $
         results <&> \(txwh,_) -> dbToApiTxSummary (txwhSummary txwh) (txwhContHistory txwh)
-searchTxs logger pool req _givenMbLim _mbOffset Nothing (Just pactid) minheight maxheight mbNext = do
-  liftIO $ logger Info $ fromString $ printf
+searchTxs logger pool req _givenMbLim _mbOffset Nothing (Just pactid) _minheight _maxheight _mbNext = liftIO $ do
+  logger Info $ fromString $ printf
     "Transaction search from %s: PactId string: %s" (show $ remoteHost req) (T.unpack pactid)
 
-  let convert = undefined
-
-  liftIO (M.with pool (queryTxsByPactId logger pactid . fst)) >>= \case
-    [] -> throw404 $ "No transactions found for pactid: " <> toS pactid
-    xs -> return $ convert xs
+  M.with pool (fmap noHeader . queryTxsByPactId logger pactid . fst)
 searchTxs _logger _pool _req _givenMbLim _mbOffset (Just _search) (Just _pactId) _minheight _maxheight _mbNext =
   throw404 "You should only specify a pactid or a search string, not both!"
 
@@ -454,6 +450,21 @@ toApiTxDetail tx contHist blk evs = TxDetail
     toTxEvent ev =
       TxEvent (_ev_qualName ev) (unPgJsonb $ _ev_params ev)
 
+toApiTxSummary :: Transaction -> ContinuationHistory -> Block -> TxSummary
+toApiTxSummary tx contHist blk = TxSummary
+        { _txSummary_chain = fromIntegral $ _tx_chainId tx
+        , _txSummary_height = fromIntegral $ _block_height blk
+        , _txSummary_blockHash = unDbHash $ unBlockId $ _tx_block tx
+        , _txSummary_creationTime = _tx_creationTime tx
+        , _txSummary_requestKey = unDbHash $ _tx_requestKey tx
+        , _txSummary_sender = _tx_sender tx
+        , _txSummary_code = _tx_code tx
+        , _txSummary_continuation = unPgJsonb <$> _tx_continuation tx
+        , _txSummary_result = maybe TxFailed (const TxSucceeded) $ _tx_goodResult tx
+        , _txSummary_initialCode = chCode contHist
+        , _txSummary_previousSteps = V.toList (chSteps contHist) <$ chCode contHist
+        }
+
 queryTxsByKey :: LogFunctionIO Text -> Text -> Connection -> IO [TxDetail]
 queryTxsByKey logger rk c =
   runBeamPostgresDebug (logger Debug . T.pack) c $ do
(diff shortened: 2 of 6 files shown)


haskell-src/exec/Chainweb/Server.hs
```diff
@@ -399,31 +399,15 @@ searchTxs logger pool req givenMbLim mbOffset (Just search) Nothing minheight ma
...
-        (runBeamPostgresDebug (logger Debug . T.pack) c)
-        (toTxSearchCursor . txwhSummary)
-        (txSearchSourcePactId pactid $ HeightRangeParams minheight maxheight)
-        noDecoration
-        continuation
-        resultLimit
-      return $ maybe noHeader (addHeader . mkTxToken) mbCont $
-        results <&> \(txwh,_) -> dbToApiTxSummary (txwhSummary txwh) (txwhContHistory txwh)
+  liftIO (M.with pool (queryTxsByPactId logger pactid . fst)) >>= \case
+    [] -> throw404 $ "No transactions found for pactid: " <> toS pactid
```
enobayram, line 409: I don't think we need to handle the empty results case specially here, returning an empty list of transactions should be fine.

emmanueldenloye:
For future onlookers, I've posted the query plan of the pactid search query before optimization. Here it is with a search parameter of `apactid`. This is not a great search paramater, however, the query plan shows that we use an inefficient query to gather transactions that match a particular pactid.
```
                                                                                     QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=1044111229.77..1044114175.56 rows=1178316 width=2336) (actual time=33506.153..33506.158 rows=0 loops=1)
   Sort Key: t0.goodresult DESC, t0.height DESC
   Sort Method: quicksort  Memory: 25kB
   ->  Nested Loop  (cost=876.90..1040428108.15 rows=1178316 width=2336) (actual time=33506.083..33506.087 rows=0 loops=1)
         ->  Nested Loop  (cost=876.34..1035573774.66 rows=1178316 width=1970) (actual time=33506.082..33506.086 rows=0 loops=1)
               ->  Index Scan using transactions_pkey on transactions t0  (cost=0.56..3588756.96 rows=1178316 width=1906) (actual time=33506.081..33506.081 rows=0 loops=1)
                     Filter: CASE WHEN (pactid IS NOT NULL) THEN ((pactid)::text = 'apactid'::text) ELSE false END
                     Rows Removed by Filter: 2356631
               ->  Aggregate  (cost=875.78..875.79 rows=1 width=64) (never executed)
                     CTE transactionsteps
                       ->  Recursive Union  (cost=0.56..873.25 rows=101 width=306) (never executed)
                             ->  Unique  (cost=0.56..8.57 rows=1 width=306) (never executed)
                                   ->  Index Scan using transactions_requestkey_idx on transactions tinner  (cost=0.56..8.57 rows=1 width=306) (never executed)
                                         Index Cond: ((requestkey)::text = ($1)::text)
                             ->  Subquery Scan on "*SELECT* 2"  (cost=86.22..86.37 rows=10 width=306) (never executed)
                                   ->  Unique  (cost=86.22..86.27 rows=10 width=310) (never executed)
                                         ->  Sort  (cost=86.22..86.24 rows=10 width=310) (never executed)
                                               Sort Key: trec.depth
                                               ->  Nested Loop  (cost=0.56..86.05 rows=10 width=310) (never executed)
                                                     ->  WorkTable Scan on transactionsteps trec  (cost=0.00..0.20 rows=10 width=36) (never executed)
                                                     ->  Index Scan using transactions_requestkey_idx on transactions tinner_1  (cost=0.56..8.57 rows=1 width=302) (never executed)
                                                           Index Cond: ((requestkey)::text = (trec.pactid)::text)
                     ->  CTE Scan on transactionsteps  (cost=0.00..2.02 rows=101 width=68) (never executed)
         ->  Index Scan using blocks_pkey on blocks t2  (cost=0.56..4.12 rows=1 width=366) (never executed)
               Index Cond: ((hash)::text = (t0.block)::text)
 Planning Time: 0.490 ms
 JIT:
   Functions: 25
   Options: Inlining true, Optimization true, Expressions true, Deforming true
   Timing: Generation 5.764 ms, Inlining 178.061 ms, Optimization 693.393 ms, Emission 192.643 ms, Total 1069.861 ms
 Execution Time: 33512.227 ms
 ```
 To optimize this query, we create this index
 ```
 CREATE INDEX transactions_pactid_index
  ON transactions (pactid, goodresult DESC, height DESC)
  WHERE pactid IS NOT NULL;
```
We also modify the search query to order its results by the values of `goodresult` and `height`. Here is the improvement after these changes. The search parameter is `zrkQoQkwClQLysRH0Ulqj92-zOYQC9x833DsEQPauFc` (something actually on chain)
```
                                                                                     QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=876.76..15995.77 rows=18 width=2336) (actual time=0.145..1.070 rows=16 loops=1)
   ->  Nested Loop  (cost=876.20..15841.37 rows=18 width=1970) (actual time=0.107..0.921 rows=16 loops=1)
         ->  Index Scan using transactions_pactid_index on transactions t0  (cost=0.42..76.73 rows=18 width=1906) (actual time=0.034..0.063 rows=16 loops=1)
               Index Cond: ((pactid)::text = 'zrkQoQkwClQLysRH0Ulqj92-zOYQC9x833DsEQPauFc'::text)
         ->  Aggregate  (cost=875.78..875.79 rows=1 width=64) (actual time=0.052..0.053 rows=1 loops=16)
               CTE transactionsteps
                 ->  Recursive Union  (cost=0.56..873.25 rows=101 width=306) (actual time=0.051..0.051 rows=0 loops=16)
                       ->  Unique  (cost=0.56..8.57 rows=1 width=306) (actual time=0.048..0.048 rows=0 loops=16)
                             ->  Index Scan using transactions_requestkey_idx on transactions tinner  (cost=0.56..8.57 rows=1 width=306) (actual time=0.047..0.047 rows=0 loops=16)
                                   Index Cond: ((requestkey)::text = ($1)::text)
                       ->  Subquery Scan on "*SELECT* 2"  (cost=86.22..86.37 rows=10 width=306) (actual time=0.003..0.003 rows=0 loops=16)
                             ->  Unique  (cost=86.22..86.27 rows=10 width=310) (actual time=0.003..0.003 rows=0 loops=16)
                                   ->  Sort  (cost=86.22..86.24 rows=10 width=310) (actual time=0.002..0.002 rows=0 loops=16)
                                         Sort Key: trec.depth
                                         Sort Method: quicksort  Memory: 25kB
                                         ->  Nested Loop  (cost=0.56..86.05 rows=10 width=310) (actual time=0.000..0.000 rows=0 loops=16)
                                               ->  WorkTable Scan on transactionsteps trec  (cost=0.00..0.20 rows=10 width=36) (actual time=0.000..0.000 rows=0 loops=16)
                                               ->  Index Scan using transactions_requestkey_idx on transactions tinner_1  (cost=0.56..8.57 rows=1 width=302) (never executed)
                                                     Index Cond: ((requestkey)::text = (trec.pactid)::text)
               ->  CTE Scan on transactionsteps  (cost=0.00..2.02 rows=101 width=68) (actual time=0.051..0.051 rows=0 loops=16)
   ->  Index Scan using blocks_pkey on blocks t2  (cost=0.56..8.58 rows=1 width=366) (actual time=0.008..0.008 rows=1 loops=16)
         Index Cond: ((hash)::text = (t0.block)::text)
 Planning Time: 2.536 ms
 Execution Time: 1.282 ms
```
Although the query times are different in the manner we'd like, it is worth pointing out that we did not use the same search parameter for the inefficient and optimized queries. However, what is important to note is that in the inefficient query, the query plan uses a index scan on primary key index (the primary key is the tuple `(requestkey,block)`) of the transactions table. Irrespective of the search parameter, scanning over this index is less efficient.

haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX transactions_pactid_index 
+  ON transactions (pactid, goodresult DESC, height DESC)
```
enobayram, line 2: Including `goodresult` in the index like this isn't very good for performance since it's an arbitrarily long text, but even more important than performance, [it will cause errors](https://dba.stackexchange.com/questions/25138/index-max-row-size-error) if we try to insert a row with `pactid IS NOT NULL` and `goodresult` is larger than a few kilobytes. I've run the following query and it turned out that the largest `goodresult` in the database is ~300kb!
``` SQL
SELECT length(goodresult::text), goodresult FROM transactions WHERE goodresult IS NOT NULL ORDER BY length(goodresult::text) DESC LIMIT 1;
```
I think we should instead index on `goodresult IS NOT NULL`

haskell-src/exec/Chainweb/Server.hs
```diff
@@ -398,6 +399,13 @@ searchTxs logger pool req givenMbLim mbOffset (Just search) minheight maxheight
         resultLimit
       return $ maybe noHeader (addHeader . mkTxToken) mbCont $
         results <&> \(txwh,_) -> dbToApiTxSummary (txwhSummary txwh) (txwhContHistory txwh)
+searchTxs logger pool req _givenMbLim _mbOffset Nothing (Just pactid) _minheight _maxheight _mbNext = liftIO $ do
+  logger Info $ fromString $ printf
+    "Transaction search from %s: PactId string: %s" (show $ remoteHost req) (T.unpack pactid)
+
+  M.with pool (fmap noHeader . queryTxsByPactId logger pactid . fst)
+searchTxs _logger _pool _req _givenMbLim _mbOffset (Just _search) (Just _pactId) _minheight _maxheight _mbNext =
+  throw404 "You should only specify a pactid or a search string, not both!"
```
enobayram, line 408: This should be a `throw400`

haskell-src/exec/Chainweb/Server.hs
```diff
@@ -442,6 +450,21 @@ toApiTxDetail tx contHist blk evs = TxDetail
...
+        , _txSummary_blockHash = unDbHash $ unBlockId $ _tx_block tx
+        , _txSummary_creationTime = _tx_creationTime tx
+        , _txSummary_requestKey = unDbHash $ _tx_requestKey tx
+        , _txSummary_sender = _tx_sender tx
+        , _txSummary_code = _tx_code tx
+        , _txSummary_continuation = unPgJsonb <$> _tx_continuation tx
+        , _txSummary_result = maybe TxFailed (const TxSucceeded) $ _tx_goodResult tx
+        , _txSummary_initialCode = chCode contHist
+        , _txSummary_previousSteps = V.toList (chSteps contHist) <$ chCode contHist
+        }
```
enobayram, line 466: I don't think you need to write this conversion function. We already have a [`toDbTxSummary`](https://github.com/kadena-io/chainweb-data/blob/f6431722672da3c9414a450fbb0dbebb7d511c7e/haskell-src/lib/ChainwebDb/Queries.hs#L88-L99) function to go from a `Transaction` to a `DbTxSummaryT`. And also a [`dbToApiTxSummary`](https://github.com/kadena-io/chainweb-data/blob/f6431722672da3c9414a450fbb0dbebb7d511c7e/haskell-src/exec/Chainweb/Server.hs#L350-L363) function to convert that to an `TxSummary`. It would be nice to use these functions because those are what the code search handler also uses, so reusing them here would reduce the chance of accidental inconsistencies.

BTW, by using `toDbTxSummary` inside your `queryTxsByPactId` `beam` query, you can also avoid copying unnecessary columns of the `Transaction` row.

haskell-src/exec/Chainweb/Server.hs
```diff
@@ -458,6 +481,25 @@ queryTxsByKey logger rk c =
...
+    r <- runSelectReturningList (queryTxsByPactId' pactid)
+    return $ (`fmap` r) $ \(tx,contHist, blk) -> toApiTxSummary tx contHist blk
+
+queryTxsByPactId' :: Text -> SqlSelect Postgres (TransactionT Identity, ContinuationHistoryT Identity, BlockT Identity)
+queryTxsByPactId' pactid =
+    select $ orderBy_ statusOrd $ do
+      tx <- all_ (_cddb_transactions database)
+      contHist <- joinContinuationHistory (_tx_pactId tx)
+      blk <- all_ (_cddb_blocks database)
+      guard_ (_tx_block tx `references_` blk)
```
enobayram, line 496: I actually don't think we need to join on the `block` row here.

haskell-src/lib/ChainwebDb/Queries.hs
```diff
@@ -159,6 +160,19 @@ txSearchSource search hgtRange = do
...
+  HeightRangeParams ->
+  Q Postgres ChainwebDataDb s (FilterMarked TxSummaryWithHistoryT (PgExpr s))
+txSearchSourcePactId pactId hgtRange = do
+  tx <- all_ $ _cddb_transactions database
+  contHist <- joinContinuationHistory (_tx_pactId tx)
+  guardInRange hgtRange (_tx_height tx)
+  let searchExp = val_ $ DbHash @TxHash pactId
+      isMatch = maybe_ (val_ False) (==. searchExp) (_tx_pactId tx)
+  return $ FilterMarked isMatch $
+    TxSummaryWithHistory (toDbTxSummary tx) contHist
```
enobayram, line 174: Doesn't seem like you're using this function anymore.

enobayram (changes requested):
I think this PR is close to completion, but I've asked for some important changes.

haskell-src/exec/Chainweb/Server.hs
emmanueldenloye: After removing the join with the blocks table, here is the resulting query plan.
```
                                                                               QUERY PLAN                                                         
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=876.78..30818.30 rows=35 width=1963) (actual time=0.124..0.126 rows=0 loops=1)
   ->  Index Scan using transactions_pactid_index on transactions t0  (cost=0.42..144.91 rows=35 width=1899) (actual time=0.124..0.124 rows=0 loops=1)
         Index Cond: ((pactid)::text = 'apactid'::text)
   ->  Aggregate  (cost=876.35..876.36 rows=1 width=64) (never executed)
         CTE transactionsteps
           ->  Recursive Union  (cost=0.56..873.82 rows=101 width=282) (never executed)
                 ->  Unique  (cost=0.56..8.57 rows=1 width=282) (never executed)
                       ->  Index Scan using transactions_requestkey_idx on transactions tinner  (cost=0.56..8.57 rows=1 width=282) (never executed)
                             Index Cond: ((requestkey)::text = ($1)::text)
                 ->  Subquery Scan on "*SELECT* 2"  (cost=86.27..86.42 rows=10 width=282) (never executed)
                       ->  Unique  (cost=86.27..86.32 rows=10 width=286) (never executed)
                             ->  Sort  (cost=86.27..86.30 rows=11 width=286) (never executed)
                                   Sort Key: trec.depth
                                   ->  Nested Loop  (cost=0.56..86.08 rows=11 width=286) (never executed)
                                         ->  WorkTable Scan on transactionsteps trec  (cost=0.00..0.20 rows=10 width=36) (never executed)
                                         ->  Index Scan using transactions_requestkey_idx on transactions tinner_1  (cost=0.56..8.57 rows=1 width=278) (never executed)
                                               Index Cond: ((requestkey)::text = (trec.pactid)::text)
         ->  CTE Scan on transactionsteps  (cost=0.00..2.02 rows=101 width=68) (never executed)
 Planning Time: 1.210 ms
 Execution Time: 0.287 ms
```

haskell-src/db-schema/migrations/2.1.1_transactions_pactid_index.sql
emmanueldenloye: While also including the later suggestion to not join on the blocks table, I've updated the index above to not include all of `goodresult` (just a predicate on it). Here is the query plan.
```
                                                                               QUERY PLAN                                                         
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=876.78..26368.20 rows=30 width=1963) (actual time=0.022..0.023 rows=0 loops=1)
   ->  Index Scan using transactions_pactid_index2 on transactions t0  (cost=0.42..76.72 rows=30 width=1899) (actual time=0.021..0.022 rows=0 loops=1)
         Index Cond: (((pactid)::text = 'zrkQoQkwClQLysRH0Ulqj92-zOYQC9x833DsEQPauFc'::text) AND ((goodresult IS NOT NULL) = true))
   ->  Aggregate  (cost=876.35..876.36 rows=1 width=64) (never executed)
         CTE transactionsteps
           ->  Recursive Union  (cost=0.56..873.82 rows=101 width=282) (never executed)
                 ->  Unique  (cost=0.56..8.57 rows=1 width=282) (never executed)
                       ->  Index Scan using transactions_requestkey_idx on transactions tinner  (cost=0.56..8.57 rows=1 width=282) (never executed)
                             Index Cond: ((requestkey)::text = ($1)::text)
                 ->  Subquery Scan on "*SELECT* 2"  (cost=86.27..86.42 rows=10 width=282) (never executed)
                       ->  Unique  (cost=86.27..86.32 rows=10 width=286) (never executed)
                             ->  Sort  (cost=86.27..86.30 rows=11 width=286) (never executed)
                                   Sort Key: trec.depth
                                   ->  Nested Loop  (cost=0.56..86.08 rows=11 width=286) (never executed)
                                         ->  WorkTable Scan on transactionsteps trec  (cost=0.00..0.20 rows=10 width=36) (never executed)
                                         ->  Index Scan using transactions_requestkey_idx on transactions tinner_1  (cost=0.56..8.57 rows=1 width=278) (never executed)
                                               Index Cond: ((requestkey)::text = (trec.pactid)::text)
         ->  CTE Scan on transactionsteps  (cost=0.00..2.02 rows=101 width=68) (never executed)
 Planning Time: 0.331 ms
 Execution Time: 0.083 ms
```
Note that we use an index `transactions_pactid_index2`, which is an index I temporarily created. It has the following expected definition.
```
 CREATE INDEX transactions_pactid_index2 ON public.transactions USING btree (pactid, ((goodresult IS NOT NULL)), height DESC) WHERE (pactid IS NOT NULL)
```

emmanueldenloye: I made a slight error in this previous query. The commit (58118184) includes a change that doesn't exclude rows with a null `goodResult` value (transactions that did not succeed), but instead we sort the results so that the successful transactions show up first. Here is the resulting query plan. Look at the above comment for an explanation of the index `transactions_pactid_index2`
```
                                                                                               QUERY PLAN                                         
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=30819.20..30819.29 rows=35 width=1964) (actual time=0.236..0.238 rows=16 loops=1)
   Sort Key: ((t0.goodresult IS NOT NULL)) DESC, t0.height DESC
   Sort Method: quicksort  Memory: 41kB
   ->  Nested Loop  (cost=876.78..30818.30 rows=35 width=1964) (actual time=0.067..0.218 rows=16 loops=1)
         ->  Index Scan using transactions_pactid_index2 on transactions t0  (cost=0.42..144.91 rows=35 width=1899) (actual time=0.019..0.028 rows=16 loops=1)
               Index Cond: ((pactid)::text = 'zrkQoQkwClQLysRH0Ulqj92-zOYQC9x833DsEQPauFc'::text)
         ->  Aggregate  (cost=876.35..876.36 rows=1 width=64) (actual time=0.011..0.011 rows=1 loops=16)
               CTE transactionsteps
                 ->  Recursive Union  (cost=0.56..873.82 rows=101 width=282) (actual time=0.005..0.008 rows=1 loops=16)
                       ->  Unique  (cost=0.56..8.57 rows=1 width=282) (actual time=0.005..0.006 rows=1 loops=16)
                             ->  Index Scan using transactions_requestkey_idx on transactions tinner  (cost=0.56..8.57 rows=1 width=282) (actual time=0.005..0.005 rows=1 loops=16)
                                   Index Cond: ((requestkey)::text = ($1)::text)
                       ->  Subquery Scan on "*SELECT* 2"  (cost=86.27..86.42 rows=10 width=282) (actual time=0.002..0.002 rows=0 loops=16)
                             ->  Unique  (cost=86.27..86.32 rows=10 width=286) (actual time=0.002..0.002 rows=0 loops=16)
                                   ->  Sort  (cost=86.27..86.30 rows=11 width=286) (actual time=0.002..0.002 rows=0 loops=16)
                                         Sort Key: trec.depth
                                         Sort Method: quicksort  Memory: 25kB
                                         ->  Nested Loop  (cost=0.56..86.08 rows=11 width=286) (actual time=0.001..0.001 rows=0 loops=16)
                                               ->  WorkTable Scan on transactionsteps trec  (cost=0.00..0.20 rows=10 width=36) (actual time=0.000..0.000 rows=1 loops=16)
                                               ->  Index Scan using transactions_requestkey_idx on transactions tinner_1  (cost=0.56..8.57 rows=1 width=278) (actual time=0.000..0.000 rows=0 loops=16)
                                                     Index Cond: ((requestkey)::text = (trec.pactid)::text)
               ->  CTE Scan on transactionsteps  (cost=0.00..2.02 rows=101 width=68) (actual time=0.006..0.008 rows=1 loops=16)
 Planning Time: 0.482 ms
 Execution Time: 0.306 ms
```

enobayram: Thanks for the information and the query plans, would you mind sharing the query itself along with the query plan. No need to format the query.

enobayram: Thanks for the query. I see 2 problems here:
1) The query doesn't have any `LIMIT` clause, seems like we're ignoring the `limit` parameter of the HTTP request and not applying any limit ourselves either, so this endpoint could return thousands of rows for certain `pactid`s.
2) Looking at the query plan, I see that Postgres is deciding to do the sorting after scanning the `transactions_pactid_index2` even though `transactions_pactid_index2` is already sorted in the same way as this query's `ORDER BY`. The trouble here is that if it needs to do the sorting after a full-scan, it will have to visit all the rows, even though it needs `LIMIT` many rows (once we have the `LIMIT` clause).

enobayram:
Here's the query (and its plan) we get after your last commit:
``` sql
=> EXPLAIN ANALYZE SELECT "t0"."chainid" AS "res0", "t0"."height" AS "res1", "t0"."block" AS "res2", "t0"."creationtime" AS "res3", "t0"."requestkey" AS "res4", "t0"."sender" AS "res5", "t0"."code" AS "res6", "t0"."continuation" AS "res7", "t0"."goodresult" AS "res8", "t1"."r0" AS "res9", "t1"."r1" AS "res10" FROM "transactions" AS "t0" CROSS JOIN LATERAL ( WITH RECURSIVE transactionSteps AS ( SELECT DISTINCT ON (depth) tInner.code, tInner.pactid, 1 AS depth, tInner.requestkey FROM transactions AS tInner WHERE (tInner.requestkey) = ("t0"."pactid")UNION ALL SELECT DISTINCT ON (depth) tInner.code, tInner.pactid, tRec.depth + 1, tInner.requestkey FROM transactions AS tInner INNER JOIN transactionSteps AS tRec ON tRec.pactid = tInner.requestkey )SELECT (array_agg(code) FILTER (WHERE code IS NOT NULL))[1] as code , array_agg(requestkey ORDER BY depth) as steps FROM transactionSteps ) "t1"("r0", "r1") WHERE ("t0"."pactid") = ('o3h1bK6-_68WEc5iMf0Cv53UXGl97y0Dljo8V9prFnA') ORDER BY ("t0"."goodresult") IS NOT NULL DESC, "t0"."height" DESC LIMIT 50;
                                                                                                 QUERY PLAN                                                                                                  
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=877.49..37065.52 rows=50 width=1191) (actual time=0.160..0.374 rows=50 loops=1)
   ->  Nested Loop  (cost=877.49..3402552.55 rows=4700 width=1191) (actual time=0.158..0.361 rows=50 loops=1)
         ->  Index Scan using transactions_pactid_index on transactions t0  (cost=0.55..19010.73 rows=4700 width=1170) (actual time=0.046..0.141 rows=50 loops=1)
               Index Cond: ((pactid)::text = 'o3h1bK6-_68WEc5iMf0Cv53UXGl97y0Dljo8V9prFnA'::text)
         ->  Memoize  (cost=876.94..876.96 rows=1 width=64) (actual time=0.003..0.003 rows=1 loops=50)
               Cache Key: t0.pactid
               Cache Mode: binary
               Hits: 49  Misses: 1  Evictions: 0  Overflows: 0  Memory Usage: 1kB
               ->  Aggregate  (cost=876.93..876.94 rows=1 width=64) (actual time=0.093..0.098 rows=1 loops=1)
                     CTE transactionsteps
                       ->  Recursive Union  (cost=0.56..874.40 rows=101 width=304) (actual time=0.038..0.062 rows=1 loops=1)
                             ->  Unique  (cost=0.56..8.58 rows=1 width=304) (actual time=0.036..0.040 rows=1 loops=1)
                                   ->  Index Scan using transactions_requestkey_idx on transactions tinner  (cost=0.56..8.58 rows=1 width=304) (actual time=0.032..0.034 rows=1 loops=1)
                                         Index Cond: ((requestkey)::text = ($1)::text)
                             ->  Subquery Scan on "*SELECT* 2"  (cost=86.32..86.48 rows=10 width=304) (actual time=0.016..0.018 rows=0 loops=1)
                                   ->  Unique  (cost=86.32..86.38 rows=10 width=308) (actual time=0.015..0.016 rows=0 loops=1)
                                         ->  Sort  (cost=86.32..86.35 rows=12 width=308) (actual time=0.013..0.015 rows=0 loops=1)
                                               Sort Key: trec.depth
                                               Sort Method: quicksort  Memory: 25kB
                                               ->  Nested Loop  (cost=0.56..86.10 rows=12 width=308) (actual time=0.007..0.008 rows=0 loops=1)
                                                     ->  WorkTable Scan on transactionsteps trec  (cost=0.00..0.20 rows=10 width=36) (actual time=0.001..0.002 rows=1 loops=1)
                                                     ->  Index Scan using transactions_requestkey_idx on transactions tinner_1  (cost=0.56..8.58 rows=1 width=300) (actual time=0.003..0.003 rows=0 loops=1)
                                                           Index Cond: ((requestkey)::text = (trec.pactid)::text)
                     ->  CTE Scan on transactionsteps  (cost=0.00..2.02 rows=101 width=68) (actual time=0.041..0.063 rows=1 loops=1)
 Planning Time: 0.834 ms
 Execution Time: 0.570 ms
(26 rows)
```
It looks great!

enobayram (approved):
LGTM!

PR #148, opened 2023-05-02
Remove server state data definition

emmanueldenloye:
The db should be handling this state and not an IORef.

* We've removed the need to track the max height in this commit.

### DIFF
--- a/haskell-src/exec/Chainweb/Listen.hs
+++ b/haskell-src/exec/Chainweb/Listen.hs
@@ -21,7 +21,7 @@ import           ChainwebData.TxSummary (mkTxSummary)
 import           ChainwebData.Types
 import           ChainwebDb.Types.DbHash (DbHash(..))
 import           Control.Exception
-import           Control.Monad (when)
+import           Control.Monad (forM_, when)
 import qualified Data.ByteString.Char8 as B
 import qualified Data.Pool as P
 import qualified Data.Sequence as S
@@ -34,7 +34,6 @@ import           Database.Beam.Postgres (Connection)
 import           Network.HTTP.Client
 import           Network.Wai.EventSource.Streaming
 import qualified Streaming.Prelude as SP
-import           System.IO (hFlush, stdout)
 import           System.Logger.Types hiding (logg)
 import           Text.Printf
 
--- a/haskell-src/exec/Chainweb/Listen.hs
+++ b/haskell-src/exec/Chainweb/Listen.hs
@@ -81,7 +81,6 @@ processNewHeader logTxSummaries env ph@(PowHeader h _) = do
         logg Info $ fromString $ msg <> addendum
         mapM_ (logg Info . fromString . show) ts
       insertNewHeader (_nodeInfo_chainwebVer $ _env_nodeInfo env) (_env_dbConnPool env) ph pl
-      logg Info (fromString $ printf "%d" (unChainId chain)) >> hFlush stdout
 
 insertNewHeader :: T.Text -> P.Pool Connection -> PowHeader -> BlockPayloadWithOutputs -> IO ()
 insertNewHeader version pool ph pl = do
--- a/haskell-src/exec/Chainweb/Listen.hs
+++ b/haskell-src/exec/Chainweb/Listen.hs
@@ -5,8 +5,8 @@
 module Chainweb.Listen
   ( listen
   , listenWithHandler
-  , getOutputsAndInsert
   , insertNewHeader
+  , processNewHeader
   ) where
 
 import           Chainweb.Api.BlockHeader (BlockHeader(..))
@@ -17,11 +17,14 @@ import           Chainweb.Api.NodeInfo
 import           ChainwebData.Env
 import           Chainweb.Lookups
 import           Chainweb.Worker
+import           ChainwebData.TxSummary (mkTxSummary)
 import           ChainwebData.Types
 import           ChainwebDb.Types.DbHash (DbHash(..))
 import           Control.Exception
+import           Control.Monad (when)
 import qualified Data.ByteString.Char8 as B
 import qualified Data.Pool as P
+import qualified Data.Sequence as S
 import           Data.String
 import qualified Data.Text as T
 import           Data.Text.Encoding
@@ -38,7 +41,7 @@ import           Text.Printf
 ---
 
 listen :: Env -> IO ()
-listen e = listenWithHandler e (getOutputsAndInsert e)
+listen e = listenWithHandler e (processNewHeader False e)
 
 -- | This function does not do any exception handling because it is used in
 -- multiple contexts.  When running via the 'listen' command we want it to fail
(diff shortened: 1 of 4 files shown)


enobayram (changes requested):
Thank you for this PR @emmanueldenloye, I see that this PR made things conceptually simpler, so I've asked a few changes to move it even further in this simplicity.

Also, would you mind making a request to the stats endpoint after this PR and share its response, timing information and the database query that gets executed.

haskell-src/exec/Chainweb/Server.hs
```diff
@@ -228,8 +228,8 @@ scheduledUpdates env pool ssRef runFill fillDelay = forever $ do
     now <- getCurrentTime
     logg Info $ fromString $ show now
     logg Info "Recalculating coins in circulation:"
-    height <- _ssHighestBlockHeight <$> readIORef ssRef
-    let circulatingCoins = getCirculatingCoins (fromIntegral height) now
+    height <- getMaxBlockHeight logg pool
+    let circulatingCoins = getCirculatingCoins (fromIntegral $ fromMaybe (error "") height) now
```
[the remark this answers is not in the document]
emmanueldenloye: I've removed it already. But thank you for noticing.

haskell-src/exec/Chainweb/Server.hs
```diff
@@ -660,11 +634,15 @@ recentTxsHandler logger pool = liftIO $ do
...
-        aggregate_ (\_ -> as_ @Int64 countAll_) (all_ (_cddb_transactions database))
+getTransactionCountEstimate :: LogFunctionIO Text -> Connection -> IO (Maybe Int64)
+getTransactionCountEstimate logger c = do
+    query_ c "SELECT reltuples::bigint AS estimate FROM pg_class WHERE relname='transactions'"
+      >>= \case
+        [Only (estimate :: Int64)] -> do
+          return $ Just estimate
+        _ -> do
+          logger Error "Could not get transaction count estimate"
+          return Nothing
```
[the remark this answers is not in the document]
enobayram: To be clear, by throwing an exception, I don't mean anything like a new exception type a caller could handle, I just mean an informative `error` call.

haskell-src/exec/Chainweb/Server.hs
```diff
@@ -660,11 +632,14 @@ recentTxsHandler logger pool = liftIO $ do
...
-      runBeamPostgresDebug (logger Debug . T.pack) c $ runSelectReturningOne $ select $
-        aggregate_ (\_ -> as_ @Int64 countAll_) (all_ (_cddb_transactions database))
+getTransactionCountEstimate :: Connection -> IO Int64
+getTransactionCountEstimate c =
+    query_ c stmt
+      >>= \case
+        [Only (estimate :: Int64)] -> return estimate
+        _ -> throwIO $ userError "Could not get transaction count estimate"
+  where
+    stmt = "SELECT reltuples::bigint AS estimate FROM pg_class WHERE relname='transactions'"
```
enobayram, line 642: I've just realized that querying the `pg_class` table like this can be fragile for `chainweb-data` operators that point their instance at a database that stores other data in it. For example, if the database has a `transactions` table in a schema other than the one `chainweb-data` writes to, this query will return multiple rows even if that schema is not on the `SEARCH_PATH` of `chainweb-data` and I believe even in the case when that table isn't even accessible to `chainweb-data`. Observe:
``` sql
=> CREATE SCHEMA some_external_schema;
CREATE SCHEMA
=> CREATE TABLE some_external_schema.transactions (a INT);
CREATE TABLE
=> SELECT reltuples::bigint AS estimate FROM pg_class WHERE relname='transactions';
 estimate 
----------
 16392690
       -1
(2 rows)
``` 
A more robust solution is to select the `transactions` table by its `oid`:
``` sql
=> SELECT reltuples::bigint AS estimate FROM pg_class WHERE oid='transactions'::regclass::oid;
 estimate 
----------
 16392690
(1 row)
```
The expression `'transactions'::regclass::oid` will pick the `transactions` table using the same search method that's used when we refer to the `transactions` table in any other query.

So, here's my suggestion:
```suggestion
    stmt = "SELECT reltuples::bigint AS estimate FROM pg_class WHERE oid='transactions'::regclass::oid"
```

emmanueldenloye:
> Thank you for this PR @emmanueldenloye, I see that this PR made things conceptually simpler, so I've asked a few changes to move it even further in this simplicity.
> 
> Also, would you mind making a request to the stats endpoint after this PR and share its response, timing information and the database query that gets executed.

I think I've taken into account all of your criticisms. Here is how fast it takes for the stats endpoint to return.
```
[emmanuel@nixos:~]$ time curl -XGET "http://18.206.81.105:8080/stats"
{"transactionCount":7413320,"coinsInCirculation":2.29262750345e8}
real	0m0.115s
user	0m0.005s
sys	0m0.005s
```
Also, here are the queries that were executed to produce this information.
```
2023-05-11T17:58:30.080Z [Debug] [] SELECT reltuples::bigint AS estimate FROM pg_class WHERE oid='transactions'::regclass::oid
2023-05-11T17:58:30.098Z [Debug] [] SELECT MAX("t0"."height") AS "res0" FROM "blocks" AS "t0"
```

haskell-src/exec/Chainweb/Listen.hs
```diff
@@ -67,8 +70,18 @@ getOutputsAndInsert env ph@(PowHeader h _) = do
...
+      let chain = _blockHeader_chainId h
+          height = _blockHeader_height h
+          hash = _blockHeader_hash h
+          tos = _blockPayloadWithOutputs_transactionsWithOutputs pl
+          ts = S.fromList $ map (\(t,tout) -> mkTxSummary chain height hash t tout) tos
+          msg = printf "Got new header on chain %d height %d" (unChainId chain) height
+          addendum = if S.null ts then "" else printf " with %d transactions" (S.length ts)
+      when logTxSummaries $ do
+        logg Info $ fromString $ msg <> addendum
+        mapM_ (logg Info . fromString . show) ts
```
enobayram, line 82: I see two behavior differences here:
1) We're logging at the `Info` level instead of the `Debug` level performed by the old `serverHeaderHandler`.
2) We used to log the full `transactionsWithOutputs` (i.e. `tos`), but this new code switched to logging `ts`, the `TxSummary`.

I think the following would bring the changes in line:
```suggestion
        logg Debug $ fromString $ msg <> addendum
        forM_ tos $ \txWithOutput -> 
          logg Debug $ fromString $ show txWithOutput
```

haskell-src/exec/Chainweb/Listen.hs
```diff
@@ -67,8 +70,18 @@ getOutputsAndInsert env ph@(PowHeader h _) = do
...
       insertNewHeader (_nodeInfo_chainwebVer $ _env_nodeInfo env) (_env_dbConnPool env) ph pl
-      logg Info (fromString $ printf "%d" (unChainId $ _blockHeader_chainId h)) >> hFlush stdout
+      logg Info (fromString $ printf "%d" (unChainId chain)) >> hFlush stdout
```
enobayram, line 84: Seems like this log line will also get printed in `server` mode and it will be redundant since we already have the `Got new header on chain %d height %d` logs in that case.

enobayram:
@emmanueldenloye thanks for the changes and also for posting the queries! I think this PR is very close to completion; We just have that minor bump with the logs.

enobayram (approved):
LGTM!

PR #149, opened 2023-05-15
Separate ETL and HTTP services

emmanueldenloye:
In this PR, our goal is to make it possible for an operator to run chainweb-data either as a server or as an
ETL service gathering data from chainweb-node. We will still allow an operator to run chainweb-data performing the aforementioned services simultaneously.

### DIFF
--- a/haskell-src/exec/Chainweb/Server.hs
+++ b/haskell-src/exec/Chainweb/Server.hs
@@ -135,18 +135,16 @@ apiServerCut env senv cutBS = do
   let circulatingCoins = getCirculatingCoins (fromIntegral curHeight) t
   logg Info $ fromString $ "Total coins in circulation: " <> show circulatingCoins
   let pool = _env_dbConnPool env
-  let mrunFill = senv ^? _Full . _2 . etlEnv_runFill <|> senv ^? _ETL . etlEnv_runFill
-  let mfillDelay = senv ^? _Full . _2 . etlEnv_fillDelay <|> senv ^? _ETL . etlEnv_fillDelay
-  case (,) <$> mrunFill <*> mfillDelay  of
-    Just (runFill, fillDelay) -> do
+  let mETL = senv ^? _Full . _2 <|> senv ^? _ETL 
+  case mETL  of
+    Just (ETLEnv runFill fillDelay) -> do
       _ <- forkIO $ scheduledUpdates env pool runFill fillDelay
       _ <- forkIO $ retryingListener env
       return ()
     Nothing -> return ()
-  let mport = senv ^? _Full . _1 . httpEnv_port <|> senv ^? _HTTP . httpEnv_port
-  let mserveSwaggerUi = senv ^? _Full . _1 . httpEnv_serveSwaggerUi <|> senv ^? _HTTP . httpEnv_serveSwaggerUi
-  case (,) <$> mport <*> mserveSwaggerUi of
-    Just (port, serveSwaggerUi) -> do
+  let mHTTP = senv ^? _Full . _1 <|> senv ^? _HTTP
+  case mHTTP of
+    Just (HTTPEnv port serveSwaggerUi) -> do
       logg Info $ fromString "Starting chainweb-data server"
       throttledPool <- do
         loadedSrc <- mkLoadedSource $ M.managed $ P.withResource pool
--- a/haskell-src/exec/Chainweb/Server.hs
+++ b/haskell-src/exec/Chainweb/Server.hs
@@ -106,15 +106,15 @@ type ApiWithNoSwaggerUI
      = TheApi
   :<|> "cwd-spec" :> Get '[PlainText] Text -- Respond with 404
 
-apiServer :: Env -> ServerEnv -> IO ()
-apiServer env senv = do
+apiServer :: Env -> WorkerEnv -> IO ()
+apiServer env wenv = do
   ecut <- queryCut env
   let logg = _env_logger env
   case ecut of
     Left e -> do
        logg Error "Error querying cut"
        logg Info $ fromString $ show e
-    Right cutBS -> apiServerCut env senv cutBS
+    Right cutBS -> apiServerCut env wenv cutBS
 
 type ConnectionWithThrottling = (Connection, Double)
 
@@ -127,22 +127,22 @@ throttlingFactor load = if loadPerCap <= 1 then 1 else 1 / loadPerCap where
   -- without any slowdown
   loadPerCap = fromInteger load / 3
 
-apiServerCut :: Env -> ServerEnv -> ByteString -> IO ()
-apiServerCut env senv cutBS = do
+apiServerCut :: Env -> WorkerEnv -> ByteString -> IO ()
+apiServerCut env wenv cutBS = do
   let curHeight = cutMaxHeight cutBS
       logg = _env_logger env
   t <- getCurrentTime
   let circulatingCoins = getCirculatingCoins (fromIntegral curHeight) t
   logg Info $ fromString $ "Total coins in circulation: " <> show circulatingCoins
   let pool = _env_dbConnPool env
-  let mETL = senv ^? _Full . _2 <|> senv ^? _ETL 
+  let mETL = wenv ^? _Full . _2 <|> wenv ^? _ETL 
   case mETL  of
     Just (ETLEnv runFill fillDelay) -> do
       _ <- forkIO $ scheduledUpdates env pool runFill fillDelay
       _ <- forkIO $ retryingListener env
       return ()
     Nothing -> return ()
-  let mHTTP = senv ^? _Full . _1 <|> senv ^? _HTTP
+  let mHTTP = wenv ^? _Full . _1 <|> wenv ^? _HTTP
   case mHTTP of
     Just (HTTPEnv port serveSwaggerUi) -> do
       logg Info $ fromString "Starting chainweb-data server"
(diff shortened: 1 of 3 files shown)


haskell-src/lib/ChainwebData/Env.hs
```diff
@@ -224,12 +234,47 @@ data FillArgs = FillArgs
...
+
+_HTTP :: Prism' WorkerEnv HTTPEnv
+_HTTP = prism' HTTP $ \case
+  HTTP x -> Just x
+  _ -> Nothing
+
+_ETL :: Prism' WorkerEnv ETLEnv
+_ETL = prism' ETL $ \case
+  ETL x -> Just x
+  _ -> Nothing
```
enobayram, line 255: What's the benefit of defining these accessors as `Prism'`s? As far as I can tell, we're only using them to read these values, so they could just be functions of type `WorkerEnv -> Maybe XXX`.

Actually, since we don't need `Prism'`s, we can make these accessors even more useful by replacing them with:
``` suggestion
getHTTPEnv :: WorkerEnv -> Maybe HTTPEnv
getHTTPEnv = \case
  Full (env,_) -> Just env
  HTTP env -> Just env
  ETL _ -> Nothing

getETLEnv :: WorkerEnv -> Maybe ETLEnv
getETLEnv = \case
  Full (_,env) -> Just env
  ETL env -> Just env
  HTTP _ -> Nothing
```

haskell-src/exec/Chainweb/Server.hs
```diff
@@ -126,43 +127,53 @@ throttlingFactor load = if loadPerCap <= 1 then 1 else 1 / loadPerCap where
...
-    if _serverEnv_serveSwaggerUi senv
-      then serve (Proxy @ApiWithSwaggerUI) (serverApp req :<|> swaggerServer) req f
-      else serve (Proxy @ApiWithNoSwaggerUI) (serverApp req :<|> noSwaggerServer) req f
+  let mETL = wenv ^? _Full . _2 <|> wenv ^? _ETL 
+  case mETL  of
+    Just (ETLEnv runFill fillDelay) -> do
+      _ <- forkIO $ scheduledUpdates env pool runFill fillDelay
+      _ <- forkIO $ retryingListener env
+      return ()
+    Nothing -> return ()
```
enobayram, line 144: Now that we've completely decoupled the ETL from the HTTP server, I think it makes more sense to start these threads outside `Server.hs`, at [this block](https://github.com/kadena-io/chainweb-data/blob/0382866dfc2ba8ac6985c9ca785aae7cb04ab9ac/haskell-src/exec/Main.hs#L85C1-L92) where we're dispatching to the `apiServer` function. Once we move this logic out of `apiServerCut`, we can make its type more precise by letting it accept `HTTPEnv` only.

haskell-src/lib/ChainwebData/Env.hs
```diff
@@ -224,12 +234,47 @@ data FillArgs = FillArgs
...
+data HTTPEnv = HTTPEnv
+ { _httpEnv_port :: Int
+ , _httpEnv_serveSwaggerUi :: Bool
+ } deriving (Eq,Ord,Show)
+
+httpEnv_port :: Lens' HTTPEnv Int
+httpEnv_port = lens _httpEnv_port $ \x y -> x { _httpEnv_port = y }
+
+httpEnv_serveSwaggerUi :: Lens' HTTPEnv Bool
+httpEnv_serveSwaggerUi = lens _httpEnv_serveSwaggerUi $ \x y -> x { _httpEnv_serveSwaggerUi = y }
```
enobayram, line 266: Do we really need these lenses? By nature, these `XXXEnv` values are meant to be read by the rest of the codebase and not really manipulated, so I think we could just use the field accessors instead of lenses.

haskell-src/lib/ChainwebData/Env.hs
```diff
@@ -224,12 +234,47 @@ data FillArgs = FillArgs
...
+data ETLEnv = ETLEnv
+ { _etlEnv_runFill :: Bool
+ , _etlEnv_fillDelay :: Maybe Int
+ } deriving (Eq,Ord,Show)
+
+etlEnv_runFill :: Lens' ETLEnv Bool
+etlEnv_runFill = lens _etlEnv_runFill $ \x y -> x { _etlEnv_runFill = y }
+
+etlEnv_fillDelay :: Lens' ETLEnv (Maybe Int)
+etlEnv_fillDelay = lens _etlEnv_fillDelay $ \x y -> x { _etlEnv_fillDelay = y }
```
enobayram, line 277: Unnecessary lenses

haskell-src/lib/ChainwebData/Env.hs
```diff
@@ -385,8 +446,12 @@ commands = hsubparser
        (progDesc "Gaps Worker - Fills in missing blocks lost during backfill or listen (DEPRECATED)"))
   <> command "single" (info singleP
        (progDesc "Single Worker - Lookup and write the blocks at a given chain/height"))
-  <> command "server" (info (Server <$> serverP)
+  <> command "server" (info (Worker <$> serverP)
        (progDesc "Serve the chainweb-data REST API (also does listen)"))
+  <> command "etl" (info (Worker <$> etlP)
+       (progDesc "ETL Worker - Fills in missing blocks and listens for new ones"))
+  <> command "http" (info (Worker <$> httpP)
+       (progDesc "HTTP Worker - Serves the chainweb-data REST API"))
```
enobayram, line 454: I wonder if we should avoid adding two new subcommands. After all, there's very little difference between `listen` and `etl` now. AFAICT, the only difference is that `etl` has the `run-fill` argument, which we could add to `listen` and `etl` has a slightly different logging format.

Similarly `http` is just `server` without the ETL in it, so maybe we should instead add an optional `--no-etl` flag to `server`, or maybe even just make the `--service` and `--p2p` flags optional and treat it as `http` when they're missing?

Either way, I think there's some room for simplification here.

enobayram (changes requested):
Thanks for this PR @emmanueldenloye! I think this will be very useful for some CW-D operators.

I think the functionality is good, but I've asked for code style improvements and raised some questions about the command line interface for exposing this new functionality.

haskell-src/lib/ChainwebData/Env.hs
```diff
@@ -370,29 +371,33 @@ singleP = Single
...
 
-httpP :: Parser WorkerEnv
-httpP = toWorkerEnv
+httpP :: Parser ServerEnv
+httpP = toServerEnv
   <$> option auto (long "port" <> metavar "INT" <> help "Port the server will listen on")
   <*> flag False True (long "serve-swagger-ui" <> internal)
   where
-    toWorkerEnv port serveSwaggerUi = HTTP (HTTPEnv port serveSwaggerUi)
+    toServerEnv port serveSwaggerUi = HTTP (HTTPEnv port serveSwaggerUi)
```
enobayram: Seems like `httpP` is not used anymore.

enobayram (approved):
This is great, I love it! I just left a comment for deleting unused code, but otherwise code LGTM!

That said, would be nice to post the `--help` output as a comment under this PR.

haskell-src/lib/ChainwebData/Env.hs
emmanueldenloye: Done

emmanueldenloye:
Here are the relevant changes the user should see when they query for help from the binary.
```
[emmanuel@nixos:~/work/chainweb-data]$ cabal run exe:chainweb-data -- listen --help
Up to date
Usage: chainweb-data listen [[-f|--run-fill] [--delay DELAY_MICROS]]
  Node Listener - Waits for new blocks and adds them to work queue

Available options:
  -f,--run-fill            Run fill operation once a day to fill gaps
  --delay DELAY_MICROS     Number of microseconds to delay between queries to
                           the node
  -h,--help                Show this help text

[emmanuel@nixos:~/work/chainweb-data]$ cabal run exe:chainweb-data -- server --help
Up to date
Usage: chainweb-data server --port INT [-f|--run-fill] [--delay DELAY_MICROS]
                            [--no-etl]
  Serve the chainweb-data REST API (also does listen)

Available options:
  --port INT               Port the server will listen on
  -f,--run-fill            Run fill operation once a day to fill gaps
  --delay DELAY_MICROS     Number of microseconds to delay between queries to
                           the node
  --no-etl                 Disable ETL
  -h,--help                Show this help text
```
Notice that in the latter, the user can specify that the etl portion of chainweb-data not be run.

haskell-src/lib/ChainwebData/Env.hs
```diff
@@ -347,12 +371,25 @@ singleP = Single
...
 
 serverP :: Parser ServerEnv
-serverP = ServerEnv
+serverP = toServerEnv
   <$> option auto (long "port" <> metavar "INT" <> help "Port the server will listen on")
   <*> flag False True (long "run-fill" <> short 'f' <> help "Run fill operation once a day to fill gaps")
   <*> delayP
   -- The OpenAPI spec is currently rudimentary and not official so we're hiding this option
   <*> flag False True (long "serve-swagger-ui" <> internal)
+  <*> flag False True (long "no-etl" <> help "Disable ETL")
```
enobayram, line 380:
```suggestion
  <*> flag False True (long "no-listen" <> help "Disable ETL")
```

enobayram (approved):
@emmanueldenloye Thank you for the changes and the help output! I think this PR is almost ready, but I'm wondering whether we should change `--no-etl` to `--no-listen`. As far as I can  tell "ETL" is a new name for what `chainweb-data` has been calling `listen` and we still refer to it that way in other places. I think we can reduce user confusion by referring to it as `listen` here as well. No need to change the references inside the code, but let's keep the CLI simpler. I've added a change suggestion comment.

PR #152, opened 2023-07-14
Attach signer information to tx(s) endpoints

enobayram:
This PR adds the `_txDetail_signers :: [Signer]` field to the `TxDetail` served by the `GET /txs/tx` and `GET /txs/txs` endpoints.

Additionally:
* It implements the necessary `ToSchema` types for the OAS3 spec.
* Adds a schema migration that swaps the order of the primary key columns of the `signers` table so that the primary key index supports efficient signer lookups.

### DIFF
--- a/haskell-src/db-schema/migrations/2.2.0.1_swap_signers_pkey.sql
+++ b/haskell-src/db-schema/migrations/2.2.0.1_swap_signers_pkey.sql
@@ -1,0 +1,2 @@
+ALTER TABLE signers DROP CONSTRAINT signers_pkey;
+ALTER TABLE cwd.signers ADD CONSTRAINT signers_pkey PRIMARY KEY (requestkey, idx);
--- a/cabal.project
+++ b/cabal.project
@@ -18,8 +18,8 @@ source-repository-package
 source-repository-package
     type: git
     location: https://github.com/kadena-io/chainweb-api.git
-    tag: b3e28d62c622ebda0d84e136ea6c995d5f97e46f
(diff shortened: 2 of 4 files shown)


haskell-src/db-schema/migrations/2.2.0.1_swap_signers_pkey.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE signers DROP CONSTRAINT signers_pkey;
```
emmanueldenloye, line 1: We should also change the primary key in `haskell-src/lib/ChainwebDb/Types/Signer.hs`

emmanueldenloye (changes requested):

haskell-src/db-schema/migrations/2.2.0.1_swap_signers_pkey.sql
enobayram: Nice catch!

enobayram: Funny enough, the Haskell code already looks like it's declaring the first column of the primary key to be `requestkey`: https://github.com/kadena-io/chainweb-data/blob/b976ae53b4584a900a63e579b8deec53c53bd4b1/haskell-src/lib/ChainwebDb/Types/Signer.hs#L77

Seems like `beam` is being very haphazard here.

emmanueldenloye (approved):
LGTM

PR #153, opened 2023-07-21
Add sigs field to tx detail responses

enobayram:
This PR is an amendment to #152, where we omitted the `sigs` field by oversight.

PR #155, opened 2023-07-24
Update 2.2.0.1_swap_signers_pkey.sql

### LINKED ISSUE #154: Migration Fails due to Expecting Signers to be in CWD
Migration fails due to "cwd" schema expected in `haskell-src/db-schema/migrations/2.2.0.1_swap_signers_pkey.sql`

```
2023-07-24T07:12:57.664Z [Info] [] Running migration: 2.2.0.1_swap_signers_pkey.sql
chainweb-data: SqlError {sqlState = "42P01", sqlExecStatus = FatalError, sqlErrorMsg = "relation \"cwd.signers\" does not exist", sqlErrorDetail = "", sqlErrorHint = ""}
```

Solution is to remove "cwd" schmea prefix.

mbwmbw1337:
Remove "cwd" schema prefix to correct migration to close #154.


enobayram:
@mbwmbw1337, thank you for the fix, this was an oversight in #153. I've confirmed that your fix works.

enobayram (approved):
LGTM!

enobayram:
Note: It's not valid to make changes to existing schema migrations, but it should be safe in this case since a) this change wasn't released yet and b) it's very unlikely for this migration to have coincidentally succeeded for anyone that runs chainweb-data from `master`.

PR #160, opened 2023-08-10
Avoid p2p dependency

enobayram:
This PR replaces all p2p endpoint requests with service endpoint requests avoiding the need to have a p2p configuration. This simplifies `chainweb-data`'s command line interface, but more importantly, it avoids depending on CW-N's p2p endpoint, which is supposed to be for peer to peer communication only and is subject to backwards-incompatible changes.

In particular, this PR moves the following requests to the service endpoint:
* [`GET <node-endpoint>/chainweb/0.0/<network-id>/cut`](https://github.com/kadena-io/chainweb-data/pull/160/files#diff-a55128f6e70aa85de943e6bbb87a99cfcf9fba62b1d9852376fdcfc3c6556681R188)
* [`GET <node-endpoint>/chainweb/0.0/<network-id>/chain/<chain-id>/payload/<block-hash>/outputs`](https://github.com/kadena-io/chainweb-data/pull/160/files#diff-a55128f6e70aa85de943e6bbb87a99cfcf9fba62b1d9852376fdcfc3c6556681R171)

### DIFF
--- a/README.org
+++ b/README.org
@@ -145,11 +145,11 @@ As a new block comes in, its chain number is printed as a single digit.
 ~server~ is just like ~listen~, but also runs an HTTP server that serves a
 few endpoints for doing common queries.
 Additionally, it can serve an OpenAPI v3 spec of the API when the hidden
-`--serve-swagger-ui` option is enabled, offering a basic interface for interacting
+~--serve-swagger-ui~ option is enabled, offering a basic interface for interacting
 with the API. This feature, however, is kept unofficial for now due to
 its rudimentary documentation.
 
-By specifying the optional `--no-listen` argument, the server can be made read-only,
+By specifying the optional ~--no-listen~ argument, the server can be made read-only,
 allowing multiple servers to serve from the same database.
 
 **** Endpoints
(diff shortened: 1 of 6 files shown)


enobayram:
I have tested this PR by running a `chainweb-data` instance from it against a `mainnet01`  and a `fast-development` node. It works without any issues in `server` and `fill` modes.

emmanueldenloye (approved):

enobayram:
In addition to testing this PR on `mainnet01` and `fast-development`, I've also double-checked that the `cut` and `payload/outputs` endpoints affected by this PR are unconditionally served on the Service API, see:
* [The payload API server is unconditionally a part of service API](https://github.com/kadena-io/chainweb-node/blob/9e673cc5cb7cca625fd932990d827f12cb9676df/src/Chainweb/RestAPI.hs#L391)
* [The cut DB is unconditionally provided to the service API](https://github.com/kadena-io/chainweb-node/blob/9e673cc5cb7cca625fd932990d827f12cb9676df/src/Chainweb/Chainweb.hs#L820) and [having a cut DB enables the cut API](https://github.com/kadena-io/chainweb-node/blob/9e673cc5cb7cca625fd932990d827f12cb9676df/src/Chainweb/RestAPI.hs#L390).

PR #163, opened 2023-08-16
Introduce --extra-migrations-folder argument

enobayram:
This PR introduces an `--extra-migrations-folder` CLI argument in order to facilitate specializing the `chainweb-data` DB schema. It also introduces a new README section explaining the recommended way of introducing such schema specializations.

### DIFF
--- a/README.org
+++ b/README.org
@@ -323,12 +323,13 @@ with the name ~2.3.0.0.N_...~ will be executed after the migrations until versio
 before any future migrations. Likewise, ~chainweb-data~ operators that run the latest commit
 from the ~master~ branch can also inject their migrations. For example, if the latest commit
 has the last migration named ~2.2.0.1_...~, then their migrations can be named ~2.2.0.1.N_...~.
-It's important to note however, that running ~chainweb-data~ from an unreleased commit of the
-~master~ branch is not supported officially and even though we aim to avoid it, we can change
+
+It's important to note, that running ~chainweb-data~ from an unreleased commit of the
+~master~ branch is **not officially supported** and even though we aim to avoid it, we can change
 new migrations of the ~master~ branch without notice, so you may have to fix your database
 manually by undoing migrations and removing ~schema_migrations~ entries.
 
 ~chainweb-data~ operators that specialize their database schema are strongly advised to review
-the incoming migrations whenever they upgrade their ~chainweb-data~ versions. This will allow
-them to detect any potential conflicts and inject new schema migrations at the right order to
+the incoming migrations **before** they upgrade their ~chainweb-data~ versions. This will allow
+them to detect any potential conflicts and insert new schema migrations to be executed at the right moment, to
 accommodate the incoming changes.
--- a/README.org
+++ b/README.org
@@ -25,6 +25,7 @@
     - [[#single][single]]
     - [[#migrate][migrate]]
     - [[#check-schema][check-schema]]
+  - [[#specializing-the-database-schema][Specializing the Database Schema]]
 
 * Overview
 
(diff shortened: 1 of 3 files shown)


README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+
 *** check-schema
 
 ~check-schema~ is used to perform a check of the ORM definitions against the DB schema.
 
 #+begin_example
   > chainweb-data check-schema --service-host=foo.chainweb.com --dbuser=joe --dbname=chainweb-data
 #+end_example
+
+** Specializing the Database Schema
```
alber70g, line 289: Just for my understanding, is this not Markdown? I see the usage of `~code~` instead of <code>`</code>. The extension is also different, `md` vs `org`. Probably a Haskell convention?

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+
+A common use case for ~chainweb-data~ is to primarily run it as an ETL process to
```
alber70g, line 291:
```suggestion
A common use case for ~chainweb-data~ is to primarily run it as a worker process to
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+populate a Postgres database with the blockchain data.  In this case, ~chainweb-data~
```
alber70g, line 292:
```suggestion
populate a Postgres database with blockchain data.  In this case, ~chainweb-data~
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+operators often want to run their schema migrations and specialize the schema according
```
alber70g, line 293:
```suggestion
operators often want to run their schema migrations and modify the schema according
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+to their own needs. Obviously, ~chainweb-data~ can not guarantee unimpeded operation
```
alber70g, line 294:
```suggestion
to their needs. Obviously, by introducing arbitrary schema changes, we can not guarantee unimpeded operation of ~chainweb-data~.
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+under arbitrary schema changes, so any node operator that wishes to specialize the
+database takes on the responsibility of ensuring that their extra schema changes
+do not interfere with the operation of ~chainweb-data~ at the time of defining those
+changes and also in the face of schema changes that will be introduced by future
+~chainweb-data~ versions.
```
alber70g, line 299:
```suggestion
Any node operator that wishes to modify the
database, takes on the responsibility of ensuring that their changes
do not interfere with the _current_ operation of ~chainweb-data~.
A node operator is also responsible for considering their changes in the face of _future_ releases of ~chainweb-data~.
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+
+That said ~chainweb-data~ does support a structure in order to help with this process.
```
alber70g, line 301:
```suggestion
~chainweb-data~ provides a way to help with this process.
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+Any version of ~chainweb-data~ comes with a set of schema migrations included in the
+binary that are applied by default to the database at migration time. These migrations
+are defined in the ~haskell-src/db-schema/migrations~ directory. It is possible to override
+these migrations by calling ~chainweb-data~ with the optional ~--migrations-folder~ argument.
+However, in order to facilitate the typical use case, which is to add migrations to the
+default set, an ~--extra-migrations-folder~ argument is also provided.
```
alber70g, line 307:
```suggestion
However, in order to add migrations to the
default set, an ~--extra-migrations-folder~ argument is provided.
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+
+The default migrations that come with ~chainweb-data~ generally have the following file
```
alber70g, line 309:
```suggestion
The default migrations that come with ~chainweb-data~ have the following file
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+name format: ~X.Y.Z.N_NAME.sql~, where ~X.Y.Z~ is the version of ~chainweb-data~ that the
```
alber70g, line 310:
```suggestion
name format: ~X.Y.Z.N_NAME.sql~. Here ~X.Y.Z~ is the version of ~chainweb-data~ after which the
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+migration was introduced right after and ~N~ is the migration number. These migrations are executed
```
alber70g, line 311:
```suggestion
migration was introduced. ~N~ is the migration number. These migrations are executed
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+in "dictionary order" if one considers the numeric values of X,Y,Z and N to be the letters.
```
alber70g, line 312:
```suggestion
in "alphabetical order" considering X,Y,Z and N to be the elements by which they are sorted.
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+The migration procedure will scan the names and the content hashes of the existing migrations
+and expect them to be coherent with the migrations provided via the ~--extra-migrations-folder~
+and ~--migrations-folder~ arguments (or the default migrations included in the binary).
```
alber70g, line 315:
```suggestion
The migration procedure will verify the migrations provided through the argument ~--extra-migrations-folder~, and check if they're coherent with the existing migrations (provided in the binary or via the ~--migrations-folder~ argument).
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+If the existing migrations are a prefix (i.e. they are in the correct order and have no gaps)
+of the expected migrations, then the remaining migrations will be applied in the same order.
```
alber70g, line 317: I think you can remove this sentence alltogether

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+
+~chainweb-data~ operators can inject their own migrations to be executed at the right time
+by taking advantage of this name-based ordering. For example, version 2.3.0 of ~chainweb-data~
+comes with ~2.2.0.1_...~ and any future version of ~chainweb-data~ will come with migrations
+that are equal to or greater than ~2.3.0.1_...~, therefore, it is guaranteed that any migration
+with the name ~2.3.0.0.N_...~ will be executed after the migrations until version 2.3.0 and
+before any future migrations. Likewise, ~chainweb-data~ operators that run the latest commit
```
alber70g, line 324:
```suggestion
By taking advantage of this alphabetical sorting, ~chainweb-data~ operators can insert custom migrations to be executed at the for them desired moment.
For example, version 2.3.0 of ~chainweb-data~ will have migrations done on top of version 2.2.0, thus having migrations named ~2.2.0.N_...~. Creating a custom migration named ~2.3.0.0.N_...~ will guarantee that it'll be executed before the new migrations that come with version 2.3.0 which are named ~2.3.0.N_...~.
Likewise, ~chainweb-data~ operators that run the latest commit
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+from the ~master~ branch can also inject their migrations. For example, if the latest commit
+has the last migration named ~2.2.0.1_...~, then their migrations can be named ~2.2.0.1.N_...~.
+It's important to note however, that running ~chainweb-data~ from an unreleased commit of the
+~master~ branch is not supported officially and even though we aim to avoid it, we can change
+new migrations of the ~master~ branch without notice, so you may have to fix your database
+manually by undoing migrations and removing ~schema_migrations~ entries.
```
alber70g, line 330:
```suggestion

It's important to note, that running ~chainweb-data~ from an unreleased commit of the
~master~ branch is **not officially supported** and even though we aim to avoid it, we can change
new migrations of the ~master~ branch without notice, so you may have to fix your database
manually by undoing migrations and removing ~schema_migrations~ entries.
```

README.org
```diff
@@ -265,10 +268,68 @@ releases. Use the ~fill~ command instead.
...
+
+~chainweb-data~ operators that specialize their database schema are strongly advised to review
+the incoming migrations whenever they upgrade their ~chainweb-data~ versions. This will allow
```
alber70g, line 333:
```suggestion
the incoming migrations **before** they upgrade their ~chainweb-data~ versions. This will allow
```

alber70g (approved):
I made some suggestions on the README part. I did not review the implementation

alber70g (commented):
I made some suggestions on the README part. I did not review the implementation

README.org
enobayram: This README is written as an Emacs Org mode document. I wouldn't say it's a Haskell convention, it's probably more of an Emacs person thing :) (which I'm not, this document predates me).

README.org
enobayram: Your version makes it sound like there's a comparison between the migrations in `--extra-migrations-folder` and `--migrations-folder`, while in reality, We compare (`--extra-migrations-folder` + `--migrations-folder`) against the existing migrations we find in the database that we connect to. One thing I wanted to convey in this paragraph is that CW-D is very picky about the consistency between desired migrations vs. existing migrations in the database. The existing migrations need to have exactly the same contents and they need to have been run in exactly the same order without any gaps. If CW-D detects any violation of any of this, it will bail out.

alber70g: Maybe we can add something along the lines of: and checks with already executed migrations in the database

enobayram: That's a good point, reading back my version, it's not clear what "existing migrations" refers to. Then how about:

```suggestion
The migration procedure will fetch the already executed migrations from the database and check them against the migrations provided through the ~--extra-migrations-folder~ and the ~--migrations-folder~ arguments.
```

README.org
enobayram: With my suggestion above, then we can also change this to:
```suggestion
If the already executed migrations are a prefix (i.e. they were run in the correct order and have no gaps or extras)
of the expected migrations, then the rest of the migrations will be executed.
```

PR #168, opened 2023-11-02
Update README.org

trendzetter:
suggest ssd storage for the node db for performance reasons


enobayram (approved):
LGTM! Thank you for the PR @trendzetter!

PR #172, opened 2023-12-05
Build chainweb-data with GHC 9.2.8

enobayram:
This PR updates the project configuration to build chainweb-data with GHC 9.2.8. I initially intended to upgrade it to 9.6.X, but that is not possible at the moment. Here are the constraints that forced me to choose 9.2.8:
* We can't stick to 8.10 anymore, it's not supported by the latest versions of HLS and it's too far behind the ecosystem in general
* We can't upgrade to any version >= 9.4, because many of our dependencies are lagging behind. Examples are `beam-automigrate`, `gargoyle`, `beam`. Not an exhaustive list since these are the ones I've spotted while iterating on the build.

At some point, we may be forced to drop dependencies.


emmanueldenloye (approved):
LGTM

PR #177, opened 2023-12-19
Fix events treatments for txs endpoint

enobayram:
In the presence of orphans, we can have multiple executions of the same transaction yielding different events every time. This PR makes sure that each execution gets associated with its own events only.

### DIFF
--- a/haskell-src/exec/Chainweb/Server.hs
+++ b/haskell-src/exec/Chainweb/Server.hs
@@ -433,9 +433,10 @@ queryTxsByKey logger rk c =
         , Api._signer_capList = caps
         }
     let sigs = Api.Sig . unSignature . _signer_sig <$> dbSigners
+        sameBlock tx ev = (unBlockId $ _tx_block tx) == (unBlockId $ _ev_block ev)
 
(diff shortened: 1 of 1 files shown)


haskell-src/exec/Chainweb/Server.hs
```diff
@@ -433,9 +433,10 @@ queryTxsByKey logger rk c =
         , Api._signer_capList = caps
         }
     let sigs = Api.Sig . unSignature . _signer_sig <$> dbSigners
+        sameBlock tx ev = (unBlockId $ _tx_block tx) == (unBlockId $ _ev_block ev)
```
emmanueldenloye, line 436: I feel like you should put the wanted events in the let binding. Like this:

```
evsInTxBlock = filter (sameBlock tx) evs
```

emmanueldenloye (commented):
Small change....not absolutely required.

haskell-src/exec/Chainweb/Server.hs
enobayram: Thanks for the suggestion, that `toApiTxDetail` call was indeed busy enough already so I've applied your suggestion.

PR #185, opened 2024-03-26
Remove unnecessary migration haskell packages

emmanueldenloye:
We've removed gargoyle, gargoyle-postgresql and beam-automigrate

Do note that schema changes need to be checked differently than before. The command CheckSchema no longer exists. Please refer to the migrations subcommand for help.


cabal.project.freeze
```diff
@@ -166,7 +166,7 @@ constraints: any.Boolean ==0.2.4,
              any.gargoyle-postgresql-nix ==0.3.0.1,
              any.generically ==0.1.1,
              any.generics-sop ==0.5.1.3,
-             any.ghc-bignum ==1.2,
+             any.ghc-bignum >=1.2,
```
enobayram, line 169: This sort of defeats the purpose of a "freeze" file :) Besides, I'm a little puzzled. We're just removing dependencies in this PR, so why was it necessary to relax this constraint?

haskell-src/exec/Main.hs
```diff
@@ -160,7 +156,9 @@ runMigrations pool logg migAction migrations = do
 
   Mg.runMigrations migAction steps pool logg
 
-  P.withResource pool $ checkTables logg False
+  -- TODO: Implemenet checkTables with beam-migrate? However, users can just
+  -- supply CheckMigrations at the commandline
+  -- P.withResource pool $ checkTables logg False
```
enobayram, line 161:
```suggestion
```

haskell-src/lib/ChainwebDb/Types/Common.hs
```diff
@@ -19,18 +19,17 @@ module ChainwebDb.Types.Common where
...
 import           Database.Beam.Backend.Types
-import           Database.Beam.Migrate
+import           Database.Beam.Postgres (Postgres)
 ------------------------------------------------------------------------------
 import           ChainwebDb.Types.DbHash
 ------------------------------------------------------------------------------
 
 data ReqKeyOrCoinbase = RKCB_RequestKey (DbHash TxHash) | RKCB_Coinbase
-  deriving (Eq, Ord, Generic)
+  deriving (Eq, Ord, Generic, HasSqlEqualityCheck Postgres)
```
enobayram, line 32: This is odd, why do we need this new instance now?

enobayram (commented):
I think we can also get rid of this line as part of this PR: https://github.com/kadena-io/chainweb-data/blob/c9b1a026bdec01c739e0763c6a9d97d968a78986/default.nix#L43

enobayram:
Thanks to this PR, seems like we can finally upgrade to GHC 9.6.3, see my successful attempt here: https://github.com/kadena-io/chainweb-data/compare/enis/ghc-963

While experimenting with the upgrade, I've noticed that we're still depending on `beam-migrate` via `beam-postgres`. 

Also note that I've encountered resistance to the upgrade from other `beam` packages: https://github.com/kadena-io/chainweb-data/commit/6d85aeab42bc0d3969804175fb92744b4f8fac81#diff-b8ed06757045fac949c89f2139a862498ad2b6d1f82c61a31e7c91d6cf0eaa70R50-R52. I've considered opening a PR to `beam` to resolve its issues and make it GHC 9.6 friendly, but I've noticed that there's already a 6-month old PR sitting there: https://github.com/haskell-beam/beam/pull/694

I hate to say it, but we might have to consider `beam` to be in abandonware status at this point and look into dropping it as a dependency, which can be fairly realistic once we extract the `chainweb-data` HTTP API as a separate package and deprecate it so that the codebase for the ETL functionality of `chainweb-data` is much more nimble.

haskell-src/lib/ChainwebDb/Types/Common.hs
emmanueldenloye: It was there before. We need it because of this code
```
exec/Chainweb/Server.hs:419:34: error:
    • No instance for (HasSqlEqualityCheck Postgres ReqKeyOrCoinbase)
        arising from a use of ‘==.’
    • In the first argument of ‘guard_’, namely
        ‘(_ev_requestkey ev ==. val_ (RKCB_RequestKey $ DbHash rk))’
      In a stmt of a 'do' block:
        guard_ (_ev_requestkey ev ==. val_ (RKCB_RequestKey $ DbHash rk))
      In the second argument of ‘($)’, namely
        ‘do ev <- all_ (_cddb_events database)
            guard_ (_ev_requestkey ev ==. val_ (RKCB_RequestKey $ DbHash rk))
            return ev’
    |
419 |        guard_ (_ev_requestkey ev ==. val_ (RKCB_RequestKey $ DbHash rk))
    |                                  ^^^
```

emmanueldenloye: What I posted here is the error that results if we get rid of the instance `HasSqlEqualityCheck`.

cabal.project.freeze
emmanueldenloye: I see your point. I went ahead and just created another one.

haskell-src/lib/ChainwebDb/Types/Common.hs
enobayram: > It was there before.

That's not true though, check out that line on `master`:

https://github.com/kadena-io/chainweb-data/blob/c9b1a026bdec01c739e0763c6a9d97d968a78986/haskell-src/lib/ChainwebDb/Types/Common.hs#L32-L33

And master still has the `_ev_requestkey ev ==. val_ (RKCB_RequestKey $ DbHash rk)` comparison:
https://github.com/kadena-io/chainweb-data/blob/c9b1a026bdec01c739e0763c6a9d97d968a78986/haskell-src/exec/Chainweb/Server.hs#L419

And yet `master` compiles without any issues.

This is no big deal BTW, I'm just trying to understand why removing a bunch of imports could possibly force us to add a new instance

enobayram: What's concerning here is whether this changes the behavior of the equality check. After all, there was no instance before, so the equality comparison was happening one way, and now that GHC is asking for this instance, it will presumably use this derived instance, but are these the same? I imagine that you're working on this branch locally by running `cabal build`, so I wonder whether you're picking up package versions that are different from those that `nix build` uses.

enobayram: Ah OK, I see it now, the instance was declared like this before: https://github.com/kadena-io/chainweb-data/pull/185/files#diff-69d14efae0583571ce896c4c6eba186937d5d7c92c448fa90817689291ddc8d9L48

Mystery solved :)

enobayram (approved):
LGTM!

PR #186, opened 2024-03-28
Remove pact as a dependency

emmanueldenloye:
We only seemed to need it for generating the richlist. This is a lot easier than I had imagined.


enobayram:
Oh wow, this is such a relief. This should relax our dependency resolution quite a bit while upgrading our GHC or Hackage index. Thank you for the PR!

BTW, would you mind checking how much further we can slim down the `cabal.project`? Now that `pact` is gone, I have a suspicion that the vast majority of that file's contents can be deleted.

enobayram:
Seems like installing `xz` and `postgresql` inside the non-Nix MacOS CI fixed the build, so let's keep the non-Nix `macos-latest` builds for now. But since we already have the Nix builds on `macos-latest` (and also `macos-m1`), let's drop support for the non-Nix `macos-latest` build next time it causes issues. If somebody needs `chainweb-data` on macos and they don't want the nix-built version for some reason, it shouldn't be too hard for them to figure out how to set up their local environment in order to build it with `cabal`.

enobayram:
I've just added `index-state: 2024-02-01T00:00:00Z` to the `cabal.project` since we're removing the `cabal.project.freeze` file. I'm hoping this will be a more future-proof way of maintaining consistent and stable builds inside and outside Nix.

With my latest commits, in addition to these CI builds, `cabal build chainweb-data` works:
* Ubuntu; Outside Nix using `cabal` and `ghc` provisioned with `ghcup`
* Ubuntu: Inside `nix develop`
* MacOS: Outside Nix using `cabal` and `ghc` provisioned with `ghcup`, installing `xz` and `postgresql` via homebrew and using the following `cabal.project.local` file:

        package *
          extra-include-dirs:
              /opt/homebrew/include
              /opt/homebrew/opt/openssl/include
          extra-lib-dirs:
              /opt/homebrew/lib
              /opt/homebrew/opt/openssl/lib

* MacOS: Inside `nix develop`.

enobayram (approved):
LGTM! Thanks for removing the `pact` dependency @emmanueldenloye, this should make `chainweb-data` much easier to update in the future.

PR #188, opened 2024-04-10
Add Verifiers to cw-data

emmanueldenloye:
Now txHandler & txsHandler returns a TxDetail with list of verifiers (in json) to an api consumer.


haskell-src/exec/Chainweb/Lookups.hs
```diff
@@ -318,6 +318,7 @@ mkTransaction b (tx,txo) = Transaction
   , _tx_continuation = PgJSONB <$> _toutContinuation txo
   , _tx_txid = fromIntegral <$> _toutTxId txo
   , _tx_numEvents = Just $ fromIntegral $ length $ _toutEvents txo
+  , _tx_verifiers = PgJSONB <$> tx ^? to (CW._transaction_cmdStr) . key "verifiers"
```
enobayram, line 321: Is that really where the `verifiers` field appears?

emmanueldenloye: The tests I've added answer this question.

PR #193, opened 2025-02-26
Fix circulating supply calculation

edmundnoble:
decodeAllocations assumes that token_payments.csv has a header row but actually the first row is a 10M token allocation, which therefore is missed in the circulating supply calculation.

Change-Id: Id0000000fedef0d50b1e62c003c4d4056d0ea603


edmundnoble:
This is incorrect. This allocation was never created in chainweb-node in the first place.

